PR #10, opened 2019-04-04
Stop copying build-config and enable debug build

julien:


build-ckeditor.sh
```diff
@@ -23,11 +23,12 @@ git clean -fdx
...
 git checkout "$VERSION"
 
-# Grab the build config.
-cp ../build-config.js dev/builder/build-config.js
-
-# Make the release build.
-dev/builder/build.sh
+# Make the debug or release build.
+if [ -n "$DEBUG" ]; then
+	dev/builder/build.sh --build-config ../../../build-config.js --leave-js-unminified
```
wincent: How about `--leave-css-unminified` as well?

julien: Yes, that's a good idea thanks I hadn't seen that option 👍

PR #17, opened 2019-05-09
WIP: Patch workflow for ckeditor

julien:
@wincent, This is a first attempt at writing a patch workflow for ckeditor ... so please don't be afraid at what you're about to see 🙈😱

- I thought adding this info in the README was the right thing to do.

- There's probably an issue with the way the patches are generated but that's why I'm asking you for a review.

- Maybe we could combine all three scripts into one and use `./scripts/patch.sh` with an argument like: `sh ./scripts/patch.sh init|create|build` ?

Tell me what you think when you can.

Thanks!

EDIT: Once I'll have made the required changes for this to work properly, I'll apply the first patch


README.md
```diff
@@ -74,3 +74,42 @@ To update CKEditor in liferay-portal:
...
+
+	From the root of the superproject's directory, run
+
+	```sh
+	sh ./scripts/build.sh
+	```
+
+8. Don't forget to add the changes and commit
+
+
```
wincent: Something is off with the syntax here because GitHub is failing to format it:

liferay-ckeditor_README_md_at_adb1de7012376546b20a5d927296ba95a4d5ab6d_·_liferay_liferay-ckeditor

scripts/path.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+
+# Delete the "patches" directory is
+[ -d "./patches" ] && rm -r patches
```
wincent: I'd suggest not deleting the whole dir but rather its contents. If you're worried about the first run when the directory is empty or may not exist yet, you could `mkdir` it.

```sh
mkdir -p patches
rm -rf patches/*
```

scripts/path.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Go into the submodule directory and
+# generate the patch files
+cd ckeditor-dev
+
+git format-patch -1 liferay -o ../patches
```
wincent: I don't think we want `-1` here but rather something that expresses "everything between the current merge base and the `HEAD` of the "liferay" branch", which would be something like:

```sh
git format-patch master...liferay -o ../patches
```

scripts/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+set -e
+
+# Make sure submodule is registered and up-to-date.
+git submodule update --init
+
+# Apply the patch
```
wincent: nit: patch -> patches

scripts/build.sh
```diff
@@ -0,0 +1,28 @@
...
+cd ckeditor-dev
+git checkout master
+git pull
+git checkout liferay
```
wincent: Branch may not exist yet when person runs this after a fresh clone, so should probably check that first...

```sh
git branch -f liferay
git checkout liferay
```

scripts/build.sh
```diff
@@ -0,0 +1,28 @@
...
+git reset --hard origin/master
+git am ../patches/*
```
wincent: Maybe something to prepare people for failure:

```sh
echo "Applying patches from \"patches/\" directory."
if ! git am ../patches/*; then
  echo "There was a problem applying patches:"
  echo
  echo "To retry manually and fix:"
  echo
  echo "  cd ckeditor-dev"
  echo "  git am --abort"
  echo "  git am ../patches/*"
  echo
  echo "Once you are happy with the result, run scripts/patch.sh to update the contents of \"patches/\"."
  exit 1
fi
```

scripts/path.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
wincent: Typo in the filename here; I think you meant "patch.sh", not "path.sh", right?

scripts/update.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+set -e
+
+# Make sure submodule is registered and up-to-date.
+git submodule update --init
+
+# Fetch remote changes.
+cd ckeditor-dev
+git fetch
```
wincent: If you only `fetch` but don't checkout then the `reset HEAD` below doesn't reset to watch you fetched (you could do that with `git reset --hard FETCH_HEAD` if you wanted).

scripts/update.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# Make sure our working copy is clean.
+git reset --hard HEAD
+git clean -fdx
+
+# Create a liferay branch for our patches
+git checkout -b liferay
```
wincent: I think this will fail if you run this script more than once.

scripts/path.sh
julien: That's exactly what I was looking for, thanks.

scripts/path.sh
julien: Yes much better.

README.md
julien: Yes, I'll fix that thanks for the catch

scripts/build.sh
julien: 👍

scripts/build.sh
julien: Yes, sounds like a better option

scripts/path.sh
julien: Yes exactly, that said, this file will go away.

julien:
@wincent, I'm happier with the result than I was yesterday but still need your help to review when you can.

Obviously, this is not foolproof (i.e. I can still delete the patches directory and try to run build which will fail at the moment) so I could probably improve a thing or two, but I think it will still be more useful than having to do everything manually.

Thanks again for the help!

julien:
> I could probably improve a thing or two

- [ ] The build script should still work as before (allowing a `$VERSION` to be passed)
- [ ] Update README.md
- [ ] Update post in loop

_Nice to have_

- [ ] More emojis?
- [ ] Small .gifs or videos to show the "tool" in action

README.md
```diff
@@ -74,3 +74,42 @@ To update CKEditor in liferay-portal:
...
+2. Set up everything to start working on a patch:
+
+	```sh
+	sh ck.sh setup
+	```
+
+3. Work on your changes:
+
+	This could be `cherry-pick`ing a previously created commit or manually editing a file, so this can't be automated.
```
wincent: I'd preface this with:

> `cd` into the `ckeditor-dev/` submodule and prepare your desired changes on the "liferay" branch.

README.md
```diff
@@ -74,3 +74,42 @@ To update CKEditor in liferay-portal:
...
+4. Create your commit, add your changes and write a good commit message.
+
+5. Navigate back to the superproject's root directory and create the patch:
+
+	```sh
+	cd ..
+	sh ck.sh patch
+	```
+
+7. Create a build of CKEditor containing the patches:
```
wincent: In the rendered output, this is "6". You could update this (and the following one) to match, or leave it. Depends on whether you care more about [minimizing effort](http://wiki.c2.com/?LazinessImpatienceHubris) or making the source Markdown as readable/correct as possible. For example, some people always just make every numbered list item "1." in Markdown.

README.md
```diff
@@ -74,3 +74,42 @@ To update CKEditor in liferay-portal:
...
+
+	From the root of the superproject's directory, run
+
+	```sh
+	sh ck.sh build
+	```
+
+8. Don't forget to add the changes and commit
+
+
```
wincent: Killable blank lines?

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+
+set -e
+
+function init() {
+	# Make sure submodule is registered and up-to-date.
+	git submodule update --init
+
+	# Fetch remote changes.
+	cd ckeditor-dev
+	# not needed git fetch
```
wincent: Deletable comment?

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+
+	# Make sure our working copy is clean.
+	git reset --hard HEAD --quiet
+	git clean -fdx
+}
+
+function message() {
+	if [[ ! -z "$1" ]] ; then
+		echo
+		printf -- '-%.0s' {1..80}; echo ""
```
wincent: #fancy

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+		printf "%*s\n" $(( ( $(echo $1 | wc -c ) + 80 ) / 2 )) "$1"
```
wincent: #fancier!

It wasn't obvious to me that this right-aligns. You could make it obvious by pulling it into a function.

```sh
function right_justify() {
  # ...
}
```

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+		printf -- '-%.0s' {1..80}; echo ""
+		echo
+
+		if [[ ! -z  "$2" ]] ; then
+			echo
+			echo "$2" | fold -w 80
```
wincent: One thing I'm wondering about is whether our Windows-based bretheren will have access to all of the things we're using in here (like `fold`, for example). I guess we'll find out soon enough (and hopefully the answer won't be rewriting everything as a Node script...)

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+}
+
+function usage() {
+	echo "Usage: ck.sh COMMAND"
+	echo
+	echo "  Where COMMAND is either:"
+	echo
+	echo "  🔧  setup: Setup everything to start working on a patch"
+	echo "  💉  patch: Generate patches "
+	echo "  🔥  build: Generate a patched version of CKEditor"
```
wincent: 🍕🌶🌭

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+	echo
+	echo
+}
+
+# Check arguments
+if [ $# -ne 1 ]; then
+	usage
+  exit 1
```
wincent: Indentation looks off here.

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+fi
+
+COMMAND=$1
+
+case "$COMMAND" in
+	build)
+		VERSION=$2
+
+		init
```
wincent: Given that `init` is destructive (it does a `git reset --hard` and a `git clean`), I think your "Are you sure?" prompt should actually happen before this.

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+		git checkout --detach HEAD --quiet
+
+		# Warn the user and prompt to continue
```
wincent: Unnecessary comment, right?

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+		message "⚠️  WARNING"
+
+		str=$(printf "%s\n" "This will generate a patched version of CKEditor" \
+			"Are you sure you want to continue?")
+
+		read -p "$str [y/n]? " yn
```
wincent: Maybe simpler than `printf` would be:

```sh
echo "This will generate a patched version of CKEditor"
read -p "Are you sure you want to continue? [y/n]" yn
```

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+				git branch -f liferay HEAD
+				git checkout liferay
+
+				echo
+				echo "Checking for existing patches"
+				echo
+
+				if ! ls ../patches/*; then
+					echo "There doesn't seem to be any patch"
+					# TODO: Probably exit?
```
wincent: Nah, because maybe there will be times when there aren't any patches (eg. if they all get upstreamed, or if other product changes make them unnecessary).

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+		sha1=`git submodule | grep ckeditor-dev | awk '{print $1}' | sed -e s/[^0-9a-f]//`
+
+		# CD into the right place
+		cd ckeditor-dev
+
+		# Check for the existence of the liferay branch in the submodule
+		if ! git rev-parse --verify liferay &>/dev/null; then
+			str=$(printf "%s"
+				"It seems that there's no *liferay* branch in the *ckeditor-dev* submodule.\n\n" \
+				"Please run *sh ./ck.sh setup* to set up.")
```
wincent: Again here, wouldn't just multiple `echo` calls be simpler and more readable?

(Edit: Ah, I see you're passing it into the `message` function, maybe you could teach that to take multiple args, one per line? What is that function going to do with a multiline message like this, given its usage of `wc -c`?)

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+			*)
+				echo
+				echo "Aborting."
+				echo
+		esac
+		;;
+	patch)
+		init
+		# Navigate back to superproject
+		cd ..
```
wincent: It might be nice to make `init` do that on its own. It is always a bit surprising when calling a function leaves you in a different working directory.

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+
+		# Save SHA1 for later
+		sha1=`git submodule | grep ckeditor-dev | awk '{print $1}' | sed -e s/[^0-9a-f]//`
+
+		# CD into the right place
```
wincent: This comment seems a bit redundant.

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+				"Please run *sh ./ck.sh setup* to set up.")
+
+			message "❌  ERROR" "$str"
+			exit 1
+		fi
+
+		git checkout liferay --quiet
+
+		# Check for existing patches
+		if ls ../patches/* &>/dev/null; then
```
wincent: If you wanted to be stricter here you could look for files named `*.patch`.

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+
+			str=$(printf "%s\n" "This will replace any existing patches...")
+
+			message "⚠️  WARNING" "$str"
+
+			echo
+			echo "$(ls ../patches/*)"
```
wincent: Isn't this the same as just?

```sh
ls ../patches/*
```

(Although I'd suggest `../patches/*.patch` instead.)

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+			echo
+
+			# Prompt the user to confirm he wants to delete existing patches
+			read -p "Are you sure you want to continue [y/n]? " yn
```
wincent: Again because `init` is destructive, we might want to do the prompting a bit earlier.

ck.sh
```diff
@@ -0,0 +1,216 @@
...
+		  "    3. Commit your changes\n" \
+		  "    4. Run \`sh ./ck.sh patch\` to generate the patches\n")
+
+		message "✅  DONE" "$str"
+		;;
+	*)
+		usage
+		;;
+esac
+
```
wincent: Looks like an extra line.

README.md
julien: Yes that's a good idea, thanks

README.md
julien: That's interesting ... good point
As for the "fix" , I habe my doubts, I can always update it since this is easy enought, but it might be better to use bullet points? For example, in two months we might discover that there is a missing step and we'll have to "update" everything again.

README.md
julien: 👍

ck.sh
julien: 👍😱

README.md
wincent: Right, that's exactly why some people write numbered lists like this in Markdown:

```markdown
1. first
1. second
1. third
```

In [the words](https://daringfireball.net/projects/markdown/syntax) of the immortal(?) John Gruber himself:

> The point is, if you want to, you can use ordinal numbers in your ordered Markdown lists, so that the numbers in your source match the numbers in your published HTML. But if you want to be lazy, you don’t have to.

I don't really know what the right answer is. In general, I like being lazy, but I always felt weird numbering things as "1.", "1.", "1." etc so I never did it.

ck.sh
```diff
@@ -113,11 +92,16 @@ case "$COMMAND" in
...
-				message "✅  DONE" "$str"
+				echo
+				echo "*---------------------------------------------------------*"
+				echo "|                        ✅ DONE                          |"
+				echo "*---------------------------------------------------------*"
+				echo
+				echo "Don't forget to commit the result!"
+				echo
+				echo "git add -A -- ckeditor"
+				echo "git commit -m 'Update CKEDITOR'"
```
wincent: Just a minor nit; what would be an example of a good commit message here? If we can show one automatically, great; otherwise we should probably leave it off (your instruction to "commit the result" is probably enough). I just don't won't the history to wind up with a bunch of messages that all say "Update CKEDITOR", because you'll have to look at the commits (which may be large) in order to figure out what that actually means.

PR #31, opened 2020-02-04
fix: make sure the build doesn't reset submodule

julien:


wincent:
Looks good to me. Maybe at another time we can look at improving the README a bit.

PR #34, opened 2020-02-05
LPS-107869 Update CKEditor to latest version (v4.13.1)

markocikos:
Resend of https://github.com/julien/liferay-ckeditor/pull/3


julien:
Hey @markocikos,

Everything looks OK, but could you please amend both commits with the usual format.
I think in this case, you could either use
`patch:` or `fix:` in before the commit message
Thanks!

PR #36, opened 2020-02-05
LPS-107869 Update CKEditor to latest version (v4.13.1)

markocikos:
@julien A followup on https://github.com/liferay/liferay-ckeditor/pull/34, with standardized commit messages.

PR #38, opened 2020-02-12
fix: force submodule to point at v4.13.1 upstream

wincent:
This is exactly the same fix as applied in f3df483857408a956d46a13b.

[The PR updating us to CKEditor v4.13.1](https://github.com/liferay/liferay-ckeditor/pull/36) did not include the submodule change.

The explanation I gave back in f3df483857408a956d46a13b was this:

> Our submodule was pointing at un unpublished rev. The reason is that we have the submodule set to ignore changes (thanks to a2bd01439), and so our `git add` command in `ck.sh` wasn't having any effect, and we ended up committing a reference to something that we only had in one of our local repos (probably a temporary "liferay" branch).
>
> This commit adds an `-f` to the `git add` invocation to prevent this from happening again, and forces the submodule to point at a published commit.

I'm going to chalk this up to user error. The commit messages in the PR were obviously made by hand; eg. "patch: LPS-107869 Update CKEditor to 4.13.1" (3daf6321723156455b). If they had been made using the `ck.sh` script, [these two lines](https://github.com/liferay/liferay-ckeditor/blob/6d6b972394f546f3bc3b8d3e6b3be97ba7295b3c/ck.sh#L265-L266) would have run, and the message would have been something like:

    Update ckeditor-dev to 4.13.1

    411985373 (tag: 4.13.1) Updated language files.


wincent:
Going to merge this because I need to publish something on top.

PR #39, opened 2020-02-12
feat: bundle scayt and wsc plugins

wincent:
Allows us to include "scayt" ("Spell Check As You Type") and "wsc" ("Web Spell Checker") without any of [this custom stuff in our build.gradle](https://github.com/liferay/liferay-portal/blob/b560da7528daad56d9bde0ee8d482f00a07e5914/modules/apps/frontend-editor/frontend-editor-ckeditor-web/build.gradle#L60-L82), as suggested [here](https://github.com/brianchandotcom/liferay-portal/pull/84546#issuecomment-585232765).

Unfortunately though, it is not as easy as you might think. You can't just add the plugins to the `build-config.js` because the build script will die with:

    Starting CKBuilder...
    Cleaning up target folder
    Exception in thread "main" org.mozilla.javascript.JavaScriptException: Plugin does not exist: scayt.

Yet, if you look at [the CKEditor docs](https://ckeditor.com/docs/ckeditor4/latest/features/spellcheck.html), they claim:

> The out-of-the-box spelling and grammar checking functionality is provided through plugins that are included in the Standard and Full presets available from the official CKEditor Download site. You can also add them to your custom build with online builder.

The plugin pages:

- https://ckeditor.com/cke4/addon/wsc
- https://ckeditor.com/cke4/addon/scayt

Both link to [this repo](https://github.com/WebSpellChecker/ckeditor-plugin-scayt) as their source, which makes me think that one of them has a busted link... In any case, this repo is infrequently updated and doesn't have tags that we could use to grab the code (if we added it to this repo as a submodule, for example).

The README there says:

> This plugin brings the multi-language SCAYT functionality into CKEditor 4. It is integrated by default starting with Standard Package of CKEditor 4.

It also says:

> If your version of CKEditor doesn’t have SCAYT built-in, you can easily add it by ...
>
> Clone/copy this repository contents in a new "plugins/scayt" folder in your CKEditor installation.
>
> Enable the "scayt" plugin in the CKEditor configuration file (config.js)

If you go to [the builder](https://ckeditor.com/cke4/builder), you will find that the "standard" and "full" packages do indeed include the plugin files, along with the corresponding entries in the build-config.js.

All of this is to say that we can't use the normal channels to get these plug-ins. We have to download the ZIP files and extract them. As all of this operates in the world of Bash anyway (our `ck.sh` script, the CKEditor `build.sh` script), let's just do the dirty work with some shell commands.

We run them on "setup" and "update".


wincent:
@julien I am a bit puzzled by what is going on with the `menubutton/plugin.js` file... If I remove `wsc` and `scayt` from the `build-config.js` and rebuild, the file does *not* come back.

wincent:
Ah, correction. It *does* come back. I just didn't notice it because it was untracked...

PR #40, opened 2020-02-13
refactor: make patches consistent

wincent:
Just a little thing that has been bugging me for a while now: our patch files were using inconsistent titles (one had a "patch:" prefix and the other didn't).

Seeing as these patches are always driven by changes needed in liferay-portal, and they always have LPS tickets associated with them, I dropped the prefix so that both start with an LPS number. I also tweaked the commit messages so that the title contains the issue title from JIRA (ie. the motivation for the change) and the body contains whatever additional explanation (ie. the details of the change) were previously in the commit message.

No impact on build structure (no code changes), so I am not going to redo the build.


wincent:
Utterly trivial, so merging this.

PR #41, opened 2020-02-13
feat: streamline release by using liferay-js-publish

wincent:
Seeing as I have to cut a new release, streamlining the process by switching to use `liferay-js-publish`. This brings two main benefits:

1. Publishing is a one-step operation; just `yarn version --new-version 4.13.1-liferay.2`.
2. We have a nice place now to run pre-publish checks, which means we can confirm that the version number matches the expected format etc.

Test plan:

Run `yarn version --new-version 24.1.0-beta.1` (ie. with a bad version number) and see it correctly refuse to proceed:

```
error: Proposed version "24.1.0-beta.1" does not match expected format (/^4.13.1-liferay\.\d+$/)
```

Run `yarn version --new-version 4.13.1-liferay.1` (ie. intentionally re-use the previous release number) and see it correctly refuse to do anything:

```
error Command failed.
Exit code: 1
Command: git
Arguments: commit -m chore: prepare v4.13.1-liferay.1 release
```

(That's actually Git refusing to produce an empty commit.)

Run `yarn version --new-version 4.13.1-liferay.2` (ie. a good version) and see it start publishing:

```
$ npx liferay-js-publish
npx: installed 9 in 1.897s


Failed to automatically publish package! ❌

Not on "master" branch: too scared to continue 😱

Please try publishing manually.
```

ie. it was working as expected and aborted because I am on a branch. It produces a commit that looks like this:

```diff
commit f91862c2c4dfda114b152aa389e33a8b1b8e9ac8 (HEAD -> wincent/streamline-releases, tag: v4.13.1-liferay.2)
Author: Greg Hurrell <greg.hurrell@liferay.com>
Date:   Thu Feb 13 11:27:35 2020 +0100

    chore: prepare v4.13.1-liferay.2 release

diff --git a/package.json b/package.json
index e0a47a80a1..ddf6eb79ee 100644
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "liferay-ckeditor",
-  "version": "4.13.1-liferay.1",
+  "version": "4.13.1-liferay.2",
   "description": "Liferay's fork of CKEditor",
   "main": "ckeditor/ckeditor.js",
   "repository": "https://github.com/liferay/liferay-ckeditor",
```


wincent:
Only gotcha I am aware of with this approach is that liferay-npm-scripts assumes the presence of a "stable" branch, and if you don't have one, the `git checkout stable` command that it runs won't work, so you have to do it manually first:

    git checkout -b stable -t master

On the other hand, I already created a stable branch in this repo, so that means if anybody in the future clones this repo, then `git checkout stable` *will* work for them automatically, because Git is smart enough to figure out what that should do.

wincent:
@julien: I'm going to merge this on and can loop back to address any feedback that you leave later.

PR #42, opened 2020-02-13
docs: apply miscellaneous updates

wincent:
Just updates made while testing prior to release.


wincent:
As it's just doc updates and I've proof-read them, going to merge. Can apply follow-ups later if any feedback comes in.

PR #43, opened 2020-02-13
fix: explicitly define files to be included in the release

wincent:
Because the 4.13.1-liferay.2 release that I just published contains only:

    ckeditor
    └── ckeditor.js

    0 directories, 1 file

If you compare that with the previous one, you see:

    ckeditor
    ...
    ├── ckeditor.js
    ....
    └── styles.js

    191 directories, 1287 files

Cause was almost certainly the addition of the `.gitignore` file in dc0f150772a112f87444a34f4, which `yarn` evidently takes as a cue to not ship anything at all other than the file indicated by the "main" field in the package.json.

Test plan: Run `npm publish --dry-run` and see that the tarball contents look a whole lot more healthy:

https://gist.github.com/wincent/97c394b6d14fe1418647c4c9088b021f

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,9 @@
   "name": "liferay-ckeditor",
   "version": "4.13.1-liferay.2",
   "description": "Liferay's fork of CKEditor",
+  "files": [
+    "ckeditor/**/*"
+  ],
   "main": "ckeditor/ckeditor.js",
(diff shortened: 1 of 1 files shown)


wincent:
See test plan; `npm publish --dry-run` looks good now, but the only way to be reeeeally sure is to actually cut a release, so I'm going to merge this and do that now.

PR #48, opened 2020-03-25
chore: Prepare for release 4.13.1-liferay.4

julien:


patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
```diff
@@ -25,5 +25,5 @@ index aa32f0de5..289684816 100644
  				return;
  
 -- 
-2.25.0
+2.17.1
```
wincent: You really should upgrade 😂

wincent:
@julien: Just a thought which I should have had when reviewing this. In the future we should merge one PR to add the patch and *then* do the release. Because otherwise we wind up with a changelog that has an entry in it like "chore: Prepare for release 4.13.1-liferay.4" and no mention of the actually interesting thing, which is that we fixed a bug.

julien:
@wincent yeah that makes sense, thanks.

patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
julien: image
:tada:

PR #53, opened 2020-03-25
refactor: improve actions set-up

wincent:
Same as what I did in: https://github.com/liferay/liferay-frontend-guidelines/pull/138

Don't need to install Yarn because it is already there, and we can catch more errors by passing `--frozen-lockfile`.


wincent:
This one is just like the other that we merged (eg. https://github.com/liferay/liferay-frontend-guidelines/pull/138), so going to merge this.

PR #56, opened 2020-04-13
fix: update bad submodule reference

wincent:
The submodule got improperly updated in [06efb34244](https://github.com/liferay/liferay-ckeditor/commit/06efb34244be6c98962e5134e206840d8b4ec280).

It's really hard to accidentally update the submodule (you have to use `git add -f`), so I am not sure how this happened. I am going to look into it and see if I can figure out why, but in the meantime, shipping a quick fix.

See 466835dd321d8e8e4672f6e76b298e7bb0021515 for a similar recent incident.


wincent:
cc @rolandpakai — merging this to hopefully unblock you. As I said in Slack, I'm going to modify CI to catch this kind of mistake in the future.

wincent:
> I am not sure how this happened

@julien gave me the idea to look back through my shell history and we found the culprit:

CleanShot 2020-04-13 at 14 35 37@2x

`git add -u` ends up adding just like `git add -f` would. When I ran `git staged` (alias for `git diff --cached`), I didn't see the changes (would need `--ignore-submodules=none` for that). If I'd have paid close attention to the `git st` (`git status`) I would have seen it, but I missed it there and in the GitHub UI too because it is pretty subtle.

Oh well, mystery solved, at least. I don't think people are likely to make the same mistake I did as `git add -u` is somewhat esoteric, and now we have [a CI check](https://github.com/liferay/liferay-ckeditor/pull/57) that will stop this from ever happening again anyway.

PR #61, opened 2020-04-15
fix: bad path in support/version.js

wincent:
Broken by: 002fd0c56d5c304a0166b2fbced0d04058e26d73


wincent:
Going to merge this because it is blocking me from cutting a release.

PR #62, opened 2020-04-15
fix: other bad path in support/version.js

wincent:
Because 5941e58ae0f25a39d7a507b5b38aa0d5af770c7a was only a partial fix.


wincent:
🤡

wincent:
Actually tested this time.

PR #65, opened 2020-06-11
feat: Add support to create and build skins | Issue #66

carloslancha:
This pr is, of course, WIP, but I need your feedback to see if this is the right direction (or the one we which we want to go).

## Purpose of the PR:
Add support to create and build custom skins so we don't need to place all the custom styling in portal, plus we need to change some icons and that only can be done via custom skins.

## Creating skins
CKEditor [recommends to use one of their base skins to create custom skins](https://ckeditor.com/docs/ckeditor4/latest/guide/skin_sdk_customize.html) so the idea is to be able to choose which one we want to use on the creation moment.

The creation process is pretty simple, we just select the base skin, copy it to our `skins` folder with the new given name and replace the old name with the new one inside all the files.

## Building skins
Following [CKEditor docs](https://ckeditor.com/docs/ckeditor4/latest/guide/skin_sdk_build.html) we need to use their `ckbuilder` to build the skin (the java one, there's no support for skins on the .sh, maybe we can send them a pr to add it).

For now I just added a script to build the selected skin and place the result inside the `ckeditor` dist folder. Probably we want to include this as part of the entire build process, but I need some conversation here so we can find the best way to proceed.


ck.sh
```diff
@@ -85,6 +85,53 @@ case "$COMMAND" in
...
 		;;
 
+	buildskin)
+		cd skins
+
+		echo "Which skin do you want to build?"
+
+		select skin in *
+		do
+			echo "you selected $skin"
```
wincent, line 95: AFAIK, if I select a bogus entry, `$skin` will be an unset or an empty string which means that we'll ed up running:

```
java -jar dev/builder/ckbuilder/2.3.2/ckbuilder.jar --build-skin ../skins/ ../ckeditor/skins/
```

So we probably need a check for valid input before proceeding.

ck.sh
```diff
@@ -85,6 +85,53 @@ case "$COMMAND" in
...
+			java -jar dev/builder/ckbuilder/2.3.2/ckbuilder.jar --build-skin ../skins/$skin ../ckeditor/skins/$skin
+
+			break
+		done
+		;;
+
+	createskin)
+		read -r -p "What is the name of the new skin? " skinName
+
+		echo "Which skin do you want to use as base?"
```
wincent, line 108: Do we need to offer this much flexibility? Won't we be building exactly one skin from one base? If this is not something that we are going to do often, then we might be able to simplify it.

ck.sh
```diff
@@ -85,6 +85,53 @@ case "$COMMAND" in
...
+
+		select opt in *
+		do
+			cd ../..
+
+			# Create skins folder if doesn't exist
+			mkdir -p skins
+
+			# Create folder for the new skin
+			mkdir skins/"$skinName"
```
wincent, line 120: This is in part what `-p` is for. Rather than invoking `mkdir` twice:

```
     -p      Create intermediate directories as required.  If this option is not specified,
             the full path prefix of each operand must already exist.  On the other hand,
             with this option specified, no error will be reported if a directory given as an
             operand already exists.  Intermediate directories are created with permission
             bits of rwxrwxrwx (0777) as modified by the current umask, plus write and search
             permission for the owner.
```

ck.sh
```diff
@@ -85,6 +85,53 @@ case "$COMMAND" in
 		esac
 		;;
 
+	buildskin)
+		cd skins
```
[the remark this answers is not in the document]
carloslancha: gotcha

ck.sh
carloslancha: Roger that

ck.sh
carloslancha: Probably we don't need it, I need to keep working on not doing more than we need 😅 I'll get rid of it.

ck.sh
carloslancha: but here we're selecting a skin from a list, how a bogus entry could be selected?

carloslancha: said nothing

wincent: You can type random crap in, as you've probably discovered.

ck.sh
wincent: To help you, you can print these out and sleep with them under your pillow:

- https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it
- https://martinfowler.com/bliki/Yagni.html
- https://wiki.c2.com/?YouArentGonnaNeedIt

Alternatively, you could read them. Another option is starting on wiki.c2.com and spending the next 10 hours following links. 😂

ck.sh
carloslancha: yep xD

carloslancha:
@wincent pr updated

ck.sh
```diff
@@ -85,6 +85,54 @@ case "$COMMAND" in
...
+
+		echo "Which skin do you want to build?"
+
+		select skin in *
+		do
+			if [[ $skin = "" ]]; then
+				echo "selected skin doesn't exist"
+				break;
```
wincent, line 97: No semis.

carloslancha:
Oooki doki

carloslancha:
@wincent anything else? :P

julien:
@carloslancha I know this is not automated yet, but I'd suggest running shellcheck on the script just to see if there are any errors - if @wincent thinks it's also a good idea

carloslancha:
While trying to build ckeditor to test the new skin I discovered that the skins has to be placed in `ckeditor-dev/skins` not in `ckeditor/skins` to make build work. So probably having a task to build only the skin has no sense, I'm moving it to the build task.

jbalsas:
Once we have our own... I think we probably don't need any other skin, at least it shouldn't be a core feature we provide, to be able to switch skins back to default CKEditor ones. 

Just saying, I have no idea if what I said simplifies this or not :D

wincent:
@jbalsas

> Once we have our own... I think we probably don't need any other skin, at least it shouldn't be a core feature we provide, to be able to switch skins back to default CKEditor ones.

This is what I was getting at with [this comment](https://github.com/liferay/liferay-ckeditor/pull/65#discussion_r439377654):

> Do we need to offer this much flexibility? Won't we be building exactly one skin from one base? If this is not something that we are going to do often, then we might be able to simplify it.

@carloslancha 

> While trying to build ckeditor to test the new skin I discovered that the skins has to be placed in `ckeditor-dev/skins` not in `ckeditor/skins` to make build work.

Integration testing for the win!

wincent:
> just-sayin

i-love-you

carloslancha:
@jbalsas @wincent do you want me to get rid of the `createskin` script then?

wincent:
It's already there, so leave it. Once the skin is in there, once we can see that it's all working and we don't have to redo things, then we can look at cleaning it up.

carloslancha:
@wincent @jbalsas I've just pushed three commits creating the new skin moono-lexicon (just a copy of moono-lisa for now), setting it as the default skin and moving the skin build process to the general one.

README.md
```diff
@@ -188,3 +194,10 @@ To update CKEditor in liferay-portal:
 3. Re-deploy the module with `gradlew clean deploy`.
 
 An example of this can be seen in [this PR](https://github.com/brianchandotcom/liferay-portal/pull/87677).
+
+### Creating and building a custom skin
+
+1. Create a new skin running `sh ck.sh createskin`.
+2. Edit the skin at `/skins/yourskin` folder.
+3. Build the sking running `sh ck.sh buildskin`.
```
wincent, line 202: sking -> skin

ck.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+CKBUILDER_VERSION="2.3.2"
```
wincent, line 5: This seems to do nothing.

Issue #71, opened 2020-06-23
Wrong styling in editable area on source editor mode


carloslancha:
![image](https://user-images.githubusercontent.com/5803434/85400465-ba907400-b558-11ea-9218-5e1720fffcb7.png)

PR #72, opened 2020-06-23
fix: styles for editable area on source editor mode | #71

carloslancha:
Fixing styles for editable area on source editor mode.

Before:
image

After
image

### DIFF
--- a/skins/moono-lexicon/mainui.css
+++ b/skins/moono-lexicon/mainui.css
@@ -185,11 +185,11 @@ Special outer level classes used in this file:
 /* Styles for the editable area */
 .cke_editable {
 	background-color: #f1f2f5; /* $light */
-	height: 100%;
+	height: calc(100% - 12px) !important;
 	margin: 0;
 	padding: 12px 16px 8px 16px;
 	position: absolute;
-	width: 100%;
+	width: calc(100% - 12px) !important;
 }
 
 .cke_editable:after {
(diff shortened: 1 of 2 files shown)


skins/moono-lexicon/mainui.css
```diff
@@ -185,11 +185,11 @@ Special outer level classes used in this file:
 /* Styles for the editable area */
 .cke_editable {
 	background-color: #f1f2f5; /* $light */
-	height: 100%;
+	height: calc(100% - 12px) !important;
 	margin: 0;
 	padding: 12px 16px 8px 16px;
 	position: absolute;
-	width: 100%;
+	width: calc(100% - 12px) !important;
```
wincent, line 192: There are a few [known issues](https://caniuse.com/#search=calc) with `calc()` on IE, but I suspect this usage is fine.

I gather @marcoscv-work already ok'd the usage of `!important` here. You might be able to avoid that by increasing the specificity in another way 🤷 (eg. `.cke_editable.cke_editable` hack), but that obviously won't cut it if you have to beat an ID or an inline style.

carloslancha: Yep, actually I wrote this with @marcoscv-work , we need to overwrite inline styling.

marcoscv-work: I totally disagree with the use of `!Important` but in this case, we have not alternative so LGTM

PR #73, opened 2020-06-25
feat: streamline CKEditor bundle

julien:
This change reduces the number of bundled plugins from 55 to 31.
All plugins that were previously bundled can be used as [`extraPlugins`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_config.html#cfg-extraPlugins)
if required.


wincent:
I have no idea if this is correct, so consider this one of those `#ITrustYou` approvals (once you've fixed the Semantic PR bot errors).

can-i-trust-you

You also had some useful info in [your comment on JIRA](https://issues.liferay.com/browse/LPS-113972?focusedCommentId=2187611&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2187611) that might be useful to include in the commit messages and/or PR description.

julien:
@wincent ... I forgot about semantic commits :facepalm:  (and I did run the `yarn:ci` script before submitting this PR) which is bad ... I wonder if we could run this bot locally somehow (although that's off topic) - I'll fix this.

@jbalsas you're understanding correctly, but let me reply to your questions and suggestions

> Produce a ckeditor.js with 31 plugins inside 

Yes, that's correct.

> Keep the other 24 plugins in the plugins folder in our published module so they can be used as extraPlugins. (or this is not needed?)

Yes, this is actually done automatically by CKBuilder, if a plugin is "bundled", it gets copied (and minified) inside the `ckeditor.js` file, but regardless of that, the plugin's source code will also be available in the "plugins" directory - at least that's what I've seen.

For example, in my change, I've "enabled" the `a11yhelp` plugin, and when I build `ckeditor.js` with `./ck.sh build`:
 
I'll have the `a11yhelp` plugin's code inside `ckeditor.js` as seen in this screenshot

image

As well as a `ay11help` directory inside `ckeditor/plugins/` 

image

This is also confusing for me, because if the plugin is bundled I don't know why it should be there as well, so if you want I can see if this can be changed, because it would produce a "smaller" package and I think we'd all be happier.

> What happens with the rest of plugins? A quick count in the plugins folder gives me 72... so what happens to them?

From what I've seen, all plugins listed [here](https://github.com/ckeditor/ckeditor4/tree/major/plugins) get copied to the `ckeditor/plugins` directory. Which means we should also be able to use plugins that aren't bundled by default (like `cloudservices` for example)

> If those missing plugins are also inside the plugins folder in our published package... what's the point in setting the ones we remove from the bundle to 0 in this PR? Why not simply delete those lines?

Correct, we could and should remove them instead of setting them to 0 - so I'll do that now

jbalsas:
> As well as a ay11help directory inside ckeditor/plugins/

What you have there is not the plugin but the dialogDefinition.

> [...] because it would produce a "smaller" package and I think we'd all be happier

I don't think this makes a big difference. We only download this when building and we do this on big machines and big connections. Shaving off some KBs here doesn't seem worth it.

wincent:
> * If those missing plugins are also inside the plugins folder in our published package... what's the point in setting the ones we remove from the bundle to `0` in this PR? Why not simply delete those lines?

...

> Correct, we could and should remove them instead of setting them to 0 - so I'll do that now

(Devil's advocate) You could argue that there's some value in having them in there as a handy manifest of "what you could turn on if you wanted to". But yeah, I am fine with deleting them.

jbalsas:
> (Devil's advocate) You could argue that there's some value in having them in there as a handy manifest of "what you could turn on if you wanted to". But yeah, I am fine with deleting them.

I'm fine keeping them, I'd just want to be consistent. Either ALL possible things are in there as a manifest (so, 72) or none of them are (so, 31). Having ones in and ones out is what confuses me :)

julien:
> What you have there is not the plugin but the dialogDefinition.

Thanks for clarifying that @jbalsas


> I'm fine keeping them, I'd just want to be consistent. Either ALL possible things are in there as a manifest (so, 72) or NONE of them are (so, 31). Having some in and some out (55) is what confuses me :)

I agree with this too, "all" or "none", so I went ahead and removed it

julien:
Ok, going to rebase and send 2 separate PRs

wincent:
should-i-trust-you

PR #74, opened 2020-06-25
feat: add support for svg icons

carloslancha:
https://issues.liferay.com/browse/IFI-1829

We need to use SVG icons as we discussed on [slack](https://liferay.slack.com/archives/G3JBR21HA/p1592996643118600)

In this pr we're generating needed icon classes for all the different states and adding required support for `rtl` and `ltr`.

icon states

Striked out ckeditor icons:
image


skins/moono-lexicon/icons/icons.json
```diff
@@ -4,36 +4,59 @@
...
-        "redo": "redo",
+        "numberedlist": {
+            "ltr": "list-ol"
+        },
+        "outdent": {
+            "ltr": "indent-less",
+            "rtr": "indent-more"
+        },
+        "redo": {
+            "ltr": "redo",
```
jbalsas, line 43: Are you sure these are directional? It kinda makes sense... but I'd love to get some reading about it :D

skins/moono-lexicon/icons/icons.json
```diff
@@ -4,36 +4,59 @@
...
-        "bulletedlist": "list-ul",
+        "bulletedlist": {
+            "ltr": "list-ul"
+        },
         "copy": "paste",
         "cut": "cut",
-        "find": "search",
+        "find": {
+            "ltr": "search",
+            "rtl": "search"
```
jbalsas, line 19: Is this mean to be a different `search`? Or if they're the same, why separate `rtl` and `ltr`?

skins/moono-lexicon/icons/icons.json
```diff
@@ -4,36 +4,59 @@
         "align-image-center": "align-image-center",
         "align-image-left": "align-image-left",
         "align-image-right": "align-image-right",
-        "anchor": "flag-full",
+        "anchor": {
+            "ltr": "flag-full"
+        },
         "blockquote": "quote-right",
```
jbalsas, line 10: Shouldn't this one be directional as well? :D

skins/moono-lexicon/icons/icons.json
```diff
@@ -4,36 +4,59 @@
         "align-image-center": "align-image-center",
         "align-image-left": "align-image-left",
         "align-image-right": "align-image-right",
-        "anchor": "flag-full",
+        "anchor": {
+            "ltr": "flag-full"
```
jbalsas, line 8: Same here... why just `ltr`? 

I just have no idea what this is supposed to do, it just feels weird :D

support/iconsClassesGenerator.js
```diff
@@ -0,0 +1,154 @@
...
+		}`;
+
+	const focusIconCSS = 
+		`${directionClass}.cke_hidpi .cke_button:not(.cke_button_disabled):focus .cke_button__${cKEditorIcon}_icon,
+		${directionClass} .cke_button:not(.cke_button_disabled):focus .cke_button__${cKEditorIcon}_icon {
+			background: url("data:image/svg+xml;charset=utf8,${encodeSvgData(svgData, focusColor)}") !important;
+		}`;
+
+		iconsCSSContent += `${activeIconCSS} ${defaultIconCSS} ${disableIconCSS} ${hoverIconCSS} ${focusIconCSS}`;
+	*/
```
jbalsas, line 151: Are we still saving all these comments? :D 

Not sure if this is something you want to keep... but if you do, one option is to create a bogus commit adding it and one removing it, so you will have it in the git history :D

ck.sh
```diff
@@ -59,7 +59,7 @@ case "$COMMAND" in
 				cd ckeditor-dev
 
 				# Convert svg icons to png
```
jbalsas, line 61: I assume this comment is no longer acurate?

carloslancha: Upsy

skins/moono-lexicon/icons/icons.json
carloslancha: CKEditor have it as directional (automatically uses `.cke-ltr` or `.cke-rtl` classes`). I assumed it was correct.

skins/moono-lexicon/icons/icons.json
carloslancha: Same, CKEditor have it as directional so we need to specify the icon for both directions. Probably we'll need to specify other icon, something like `search-rtl`

skins/moono-lexicon/icons/icons.json
carloslancha: It is but we don't have the rtl icon yet. I can use the same until we have it or just leave it blank.

skins/moono-lexicon/icons/icons.json
carloslancha: Same as the others

support/iconsClassesGenerator.js
carloslancha: Ups, no, I have it already in the history, just forgot to remove it, thx

carloslancha:
About `rtl` icons:

CKEditor automatically adds to the css the classes for all the icons in `ltr` (just using `.cke_button .cke_button__name_icon`) and some for `rtl`(adding `.cke_rtl .cke_button .cke_button_name_icon`).

There are some of those `rlt` icons that maybe we don't want to mirror. [Materail UI] for example has some guidelines about it.

- Search icon: Lens handle is usually at the right size because majority of users are right-handed, also in RTL-writing countries.

- Undo / Redo: These icons refers to time, clockwise (right -> redo) and counterclockwise (left > undo). In Google's case they use a circular icon that could be associated better with the clock, our icons are not exactly circular but I think that still could be associated in the same way.

- Anchor / Flag: Nothing about this but I guess we can follow the same principle as the search icon. (Already listed as rtl in our [icons excel](https://docs.google.com/spreadsheets/d/1EnNDSzbKGdmnCmqWikRSBF69daUG5PKOqtzbdGolqoM/edit#gid=0))

Anyway, not an expert, maybe @drakonux can provide more info :)

carloslancha:
Forgot to add that even if we don't want to mirror some of those icons we need to specify the `rtl` icon using the same one as `ltr` just because CKEditor works in that way.

jbalsas:
LGTM, @carloslancha!

Just one question... could you highlight which of the icons in the gif are ours and which ones are ckeditor's?

carloslancha:
@jbalsas updated description

support/iconsClassesGenerator.js
```diff
@@ -0,0 +1,116 @@
...
+const getCSS = (svgData, cKEditorIcon, direction) => {
+	let directionClass = direction ? `.cke_${direction}` : '';
+
+	const activeIconCSS = `${directionClass}.cke_hidpi .cke_button.cke_button_on .cke_button__${cKEditorIcon}_icon,
+		${directionClass} .cke_button.cke_button_on .cke_button__${cKEditorIcon}_icon {
+			background: url("data:image/svg+xml;charset=utf8,${encodeSvgData(
+				svgData,
+				activeColor
+			)}") !important;
+		}`;
```
wincent, line 45: nit: I'd format these so that the *output* is consistently indented:

```
	const activeIconCSS = `
		${directionClass}.cke_hidpi .cke_button.cke_button_on .cke_button__${cKEditorIcon}_icon,
		${directionClass} .cke_button.cke_button_on .cke_button__${cKEditorIcon}_icon {
			background: url("data:image/svg+xml;charset=utf8,${encodeSvgData(
				svgData,
				activeColor
			)}") !important;
		}
	`;
```

(The way you have it, you have 0 indent on first line, then it jumps to two tabs of indent)

support/iconsClassesGenerator.js
```diff
@@ -0,0 +1,116 @@
...
+
+	const focusIconCSS = `${directionClass}.cke_hidpi .cke_button:not(.cke_button_disabled):focus .cke_button__${cKEditorIcon}_icon,
+		${directionClass} .cke_button:not(.cke_button_disabled):focus .cke_button__${cKEditorIcon}_icon {
+			background: url("data:image/svg+xml;charset=utf8,${encodeSvgData(
+				svgData,
+				focusColor
+			)}") !important;
+		}`;
+
+	return `${activeIconCSS} ${defaultIconCSS} ${disableIconCSS} ${hoverIconCSS} ${focusIconCSS}`;
```
wincent, line 79: Maybe?

```js
return [
    activeIconCSS,
    defaultIconCSS,
    disableIconCSS,
    hoverIconCSS,
    focusIconCSS,
].join('\n\n');
```

support/iconsClassesGenerator.js
```diff
@@ -0,0 +1,116 @@
...
+const hoverColor = '#272833';
+
+const encodeSvgData = (svgData, color) => {
+	const symbols = /[\r\n%#()<>?\[\\\]^`{|}]/g;
+
+	return svgData
+		.replace(/(<svg[^>]*)( fill="[^"]+")/, '\1')
+		.replace(/<svg/, `<svg fill="${color}"`)
+		.replace(/"/g, "'") // Use single quotes instead of double to avoid encoding.
+		.replace(/>\s{1,}</g, '><')
```
wincent, line 30: `{1,}` is just `+`, isn't it? Why not use that?

ck.sh
```diff
@@ -58,8 +58,8 @@ case "$COMMAND" in
 
 				cd ckeditor-dev
 
-				# Convert svg icons to png
-				node ../support/svgToPng.js skins/moono-lexicon/icons/icons.json skins/moono-lexicon/icons
+				# Generate svg icons CSS Classes
```
wincent, line 61: nit: Consistent capitalization? (ie. "svg" vs "CSS")

carloslancha:
@wincent formated

PR #83, opened 2020-07-01
docs: update stale reference to non-existent "buildskin" subcommand

wincent:
There is no "buildskin" subcommand any more, because we folded that into the "build" subcommand.


wincent:
I'm just going to merge this because it's trivial and I am going to have a merge conflict with myself if I don't...

carloslancha:
image

PR #87, opened 2020-07-02
docs: remove instruction to "git commit" after building

wincent:
Seeing as it happens automatically since 43074c10c7.

I previously tried to update the docs to describe the new behavior in 415322d2ea, but I missed this reference.


wincent:
Trivs, so going to merge.

PR #89, opened 2020-07-06
feat: allow additional resource URL parameters

### LINKED ISSUE #88: Resource URLs should contain a `languageId` parameter
Currently,  `CKEDITOR` generates URLs for it's resources (plugins and assets) with a [`t`](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/core/ckeditor_base.js#L179-L180) parameter that contains a "timestamp" (which has the value of `CKEDITOR.timestamp`), to avoid caching. In `liferay-portal` it would be convenient to add an additional `languageId` parameter so that the resources are translated.

A previous pull request has been submitted [here](https://github.com/liferay-frontend/liferay-portal/pull/77), but the right place to do it here.
This will probably need to be done via patching the existing [`getUrl`](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/core/ckeditor_base.js#L149-L183) function.

julien:
Closes #88

### DIFF
--- a/patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
+++ b/patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
@@ -1,0 +1,71 @@
+From 498269a5dd72249555f905022e5a5953410e04a3 Mon Sep 17 00:00:00 2001
+From: Julien Castelain <julien.castelain@liferay.com>
+Date: Tue, 7 Jul 2020 09:47:27 +0200
+Subject: [PATCH 5/5] LPS-112982 Add additional resource URL parameters
+
+---
+ core/ckeditor_base.js | 25 ++++++++++++++++++-------
+ core/config.js        |  9 +++++++++
+ 2 files changed, 27 insertions(+), 7 deletions(-)
+
+diff --git a/core/ckeditor_base.js b/core/ckeditor_base.js
+index 3fb18f5f7..a6ce37594 100644
+--- a/core/ckeditor_base.js
++++ b/core/ckeditor_base.js
+@@ -171,15 +171,26 @@ if ( !window.CKEDITOR ) {
+ 			 * @returns {String} The full URL.
+ 			 */
+ 			getUrl: function( resource ) {
+-				// If this is not a full or absolute path.
+-				if ( resource.indexOf( ':/' ) == -1 && resource.indexOf( '/' ) !== 0 )
+-					resource = this.basePath + resource;
++				var resourceUrl = resource.startsWith(this.basePath)
++					? new URL(resource)
++					: new URL(this.basePath + resource);
+ 
+-				// Add the timestamp, except for directories.
+-				if ( this.timestamp && resource.charAt( resource.length - 1 ) != '/' && !( /[&?]t=/ ).test( resource ) )
+-					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++				if (!resourceUrl.toString().endsWith('/')) {
++					if (this.timestamp && !resourceUrl.searchParams.has('t')) {
++						resourceUrl.searchParams.set('t', this.timestamp);
++					}
++
++					Object.entries(CKEDITOR.ADDITIONAL_RESOURCE_PARAMS).forEach(function(entries) {
++						var key = entries[0];
++						var value = entries[1];
++
++						if (!resourceUrl.searchParams.has(key)) {
++							resourceUrl.searchParams.set(key, value);
++						}
++					});
++				}
+ 
+-				return resource;
++				return resourceUrl.toString();
+ 			},
+ 
+ 			/**
+diff --git a/core/config.js b/core/config.js
+index 8a1ddb6e4..966d798d3 100644
+--- a/core/config.js
++++ b/core/config.js
+@@ -8,6 +8,15 @@
+  * default configuration settings.
+  */
+ 
++/**
++ * A map of optional additional parameters that will be
++ * added to a resource's URL
++ *
++ * @property {Object}
++ * @member CKEDITOR
++ */
++CKEDITOR.ADDITIONAL_RESOURCE_PARAMS = {};
++
+ /**
+  * Used in conjunction with the {@link CKEDITOR.config#enterMode}
+  * and {@link CKEDITOR.config#shiftEnterMode} configuration
+-- 
+2.27.0
+
--- a/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
+++ b/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
@@ -1,7 +1,7 @@
-From d407b71ca6273935cbef3304bdf1c68b59ee4c5f Mon Sep 17 00:00:00 2001
+From 8bb123947bf25d742f076fcd9facb94daa03fb27 Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Tue, 14 May 2019 10:47:25 +0200
-Subject: [PATCH 1/4] LPS-89596 Cannot Drag Image from top content line in IE11
+Subject: [PATCH 1/5] LPS-89596 Cannot Drag Image from top content line in IE11
  using Alloy Editor
 
 Changed position of drag icon for IE.
@@ -10,10 +10,10 @@ Changed position of drag icon for IE.
  1 file changed, 4 insertions(+)
 
 diff --git a/plugins/widget/plugin.js b/plugins/widget/plugin.js
-index aa32f0de5..289684816 100644
+index 1d4e036fc..a53a4c809 100644
 --- a/plugins/widget/plugin.js
 +++ b/plugins/widget/plugin.js
-@@ -1566,6 +1566,10 @@
+@@ -1607,6 +1607,10 @@
  					y: domElement.offsetTop - DRAG_HANDLER_SIZE
  				};
  
@@ -25,5 +25,5 @@ index aa32f0de5..289684816 100644
  				return;
  
 -- 
-2.25.1
+2.27.0
 
--- a/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
+++ b/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
@@ -1,7 +1,7 @@
-From d9c95371d80f904b34650b0401b220462620ffe0 Mon Sep 17 00:00:00 2001
+From 7f7f913b44304631f18fa2e9ecac7784adf77452 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Roland=20P=C3=A1kai?= <roland.pakai@liferay.com>
 Date: Tue, 21 May 2019 09:38:15 +0200
-Subject: [PATCH 2/4] LPS-95472 Tabs in popups not appears correctly in
+Subject: [PATCH 2/5] LPS-95472 Tabs in popups not appears correctly in
  maximized CKEditor
 
 Use `width: inherit` instead of `width: 100%` to work with the maximize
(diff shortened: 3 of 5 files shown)


patches/0005-LPS-112982-Add-a-languageId-parameter-to-resource-UR.patch
```diff
@@ -0,0 +1,45 @@
...
+--- a/core/ckeditor_base.js
++++ b/core/ckeditor_base.js
+@@ -171,15 +171,21 @@ if ( !window.CKEDITOR ) {
+ 			 * @returns {String} The full URL.
+ 			 */
+ 			getUrl: function( resource ) {
+-				// If this is not a full or absolute path.
+-				if ( resource.indexOf( ':/' ) == -1 && resource.indexOf( '/' ) !== 0 )
+-					resource = this.basePath + resource;
++				var resourceUrl = resource.includes(CKEDITOR.basePath)
```
wincent, line 21: Would this be clearer as `startsWith(CKEDITOR.basePath)`? (I don't know; I'm assuming it is an absolute URL.)

patches/0005-LPS-112982-Add-a-languageId-parameter-to-resource-UR.patch
```diff
@@ -0,0 +1,45 @@
...
+-					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++				if (!resourceUrl.toString().endsWith('/')) {
++					if (!resourceUrl.searchParams.has('t')) {
++						resourceUrl.searchParams.set('t', CKEDITOR.timestamp);
++					}
++
++					if (window.themeDisplay && !resourceUrl.searchParams.has('languageId')) {
++						resourceUrl.searchParams.set('languageId', window.themeDisplay.getLanguageId());
++					}
++				}
```
wincent, line 36: It looks like you dropped the `this.timestamp` check; I presume that was there for a reason though and should probably be added back, right?

Other than that, looks fine to me.

wincent: Somehow GitHub ate my other comment here 😢 — in any case, it was that you'd changed from `this.timestamp` to `CKEDITOR.timestamp`; is that the same thing?

patches/0005-LPS-112982-Add-a-languageId-parameter-to-resource-UR.patch
```diff
@@ -0,0 +1,45 @@
...
+ 
+-				// Add the timestamp, except for directories.
+-				if ( this.timestamp && resource.charAt( resource.length - 1 ) != '/' && !( /[&?]t=/ ).test( resource ) )
+-					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++				if (!resourceUrl.toString().endsWith('/')) {
++					if (!resourceUrl.searchParams.has('t')) {
++						resourceUrl.searchParams.set('t', CKEDITOR.timestamp);
++					}
++
++					if (window.themeDisplay && !resourceUrl.searchParams.has('languageId')) {
```
jbalsas, line 33: Let's not use Liferay™ APIs in here and create one that we can then push upstream.

This should rely on a similar mechanism existing in CKEditor already... off the top of my head: 

```javascript
// static param
CKEDITOR.ADDITIONAL_RESOURCE_PARAMS = {...};
```

or 

```javascript
// editor config
CKEditor.editable({
    additionalResourceURLParams: {...}
});
```

Then in DXP we could do something like:

```
CKEDITOR.ADDITIONAL_RESOURCE_PARAMS = {
    languageId: themeDisplay.getLanguageId()
};

CKEditor.editable({
    additionalResourceURLParams: {
        lanaguageId: themeDisplay.getLanguageId()
    }
});
```

patches/0005-LPS-112982-Add-a-languageId-parameter-to-resource-UR.patch
julien: Yes, in this function, `this` is `CKEDITOR`,

The parameter for `CKEDITOR.timestamp` is `t`, i.e. the URL would look something like this
`http://localhost:8000/ckeditor/styles.js?t=K669`

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
```diff
@@ -0,0 +1,71 @@
...
+ 
+-				// Add the timestamp, except for directories.
+-				if ( this.timestamp && resource.charAt( resource.length - 1 ) != '/' && !( /[&?]t=/ ).test( resource ) )
+-					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++				if (!resourceUrl.toString().endsWith('/')) {
++					if (!resourceUrl.searchParams.has('t')) {
++						resourceUrl.searchParams.set('t', CKEDITOR.timestamp);
++					}
++
++					Object.entries(CKEDITOR.ADDITIONAL_RESOURCE_PARAMS).forEach(function(entries) {
```
jbalsas, line 34: What if `CKEDITOR.ADDITIONAL_RESOURCE_PARAMS` is `null`?

jbalsas: Also, I think you can do, `function([key, value])`

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
```diff
@@ -0,0 +1,71 @@
...
+index 8a1ddb6e4..ee09ca5a6 100644
+--- a/core/config.js
++++ b/core/config.js
+@@ -8,6 +8,15 @@
+  * default configuration settings.
+  */
+ 
++/**
++ * A map of option additional parameters that will be
++ * added a resource's URL
```
wincent, line 59: "option" → "optional"

"added" → "added to"

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
jbalsas: Also, the reason I offered 2 options in my previous comment is to make sure we don't run into any race conditions where we set `CKEDITOR.ADDITIONAL_RESOURCE_PARAMS` **AFTER** we create URLs.

Can you confirm that won't be the case here or provide some insight into why you chose this option?

julien: It's always defined as an empty object, but if someone overrides it by doing something like 

`CKEDITOR.ADDITIONAL_RESOURCE_PARAMS = null;`

or something similar, it would be better to check :+1:

wincent: > Also, I think you can do, `function([key, value])`

Sure about that? This won't be transpiled, so IE [would choke on it](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment).

Still ok to use `Object.entries` though due to polyfill in DXP.

jbalsas: Ugh, forget about it, then...

Also, I didn't see https://github.com/liferay/liferay-ckeditor/pull/89/files#diff-22c7e98f6759e5b24ef2a7f9298289aaR64

I'll leave the reviews to you, @wincent!

julien: Yes, unfortunately, we can't use destructuring, but it would have been nice

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
```diff
@@ -0,0 +1,71 @@
...
++				var resourceUrl = resource.includes(CKEDITOR.basePath)
++						? new URL(resource)
++						: new URL(CKEDITOR.basePath + resource);
+ 
+-				// Add the timestamp, except for directories.
+-				if ( this.timestamp && resource.charAt( resource.length - 1 ) != '/' && !( /[&?]t=/ ).test( resource ) )
+-					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++				if (!resourceUrl.toString().endsWith('/')) {
++					if (!resourceUrl.searchParams.has('t')) {
++						resourceUrl.searchParams.set('t', CKEDITOR.timestamp);
```
wincent, line 31: Just to repeat that comment from the earlier version of this commit which might have gotten lost: old code only did this after checking that `this.timestamp` was non-falsey. I would imagine that we should maintain that check, or you could explain the reason for not maintaining it.

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
wincent: > I'll leave the reviews to you, @wincent!

But, but... [Linus's Law](https://en.wikipedia.org/wiki/Linus%27s_law)!

> given enough eyeballs, all bugs are shallow

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
julien: Yes I didn't understand all of it, we do need to check that `this.timestamp` is not falsy

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
julien: > Also, I didn't see https://github.com/liferay/liferay-ckeditor/pull/89/files#diff-22c7e98f6759e5b24ef2a7f9298289aaR64

No problem, it's actually a good idea, because even though I don't know why someone would do it, they still can do 

`CKEDITOR.ADDITIONAL_RESOURCE_PARAMS = null;`

... but should we cover all cases?
 
What if someone does this

`CKEDITOR.ADDITIONAL_RESOURCE_PARAMS = "gotcha";`

We'd still need to check that `CKEDITOR.ADDITIONAL_RESOURCE_PARAMS` is non null, and is an object, but since I know we don't want to be defensive, I'm thinking that if someone changes this to `null` or anything else than an object and it blows up they kind of get what they asked for ... but maybe that's a bit harsh

patches/0005-LPS-112982-Add-additional-resource-URL-parameters.patch
```diff
@@ -0,0 +1,73 @@
...
+ 
+-				// Add the timestamp, except for directories.
+-				if ( this.timestamp && resource.charAt( resource.length - 1 ) != '/' && !( /[&?]t=/ ).test( resource ) )
+-					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++				if (!resourceUrl.toString().endsWith('/')) {
++					if (this.timestamp && !resourceUrl.searchParams.has('t')) {
++						resourceUrl.searchParams.set('t', this.timestamp);
++					}
++
++					if (CKEDITOR.ADDITIONAL_RESOURCE_PARAMS && typeof CKEDITOR.ADDITIONAL_RESOURCE_PARAMS === 'object') {
```
wincent, line 34: `typeof null` is `'object'`, and `Object.entries(null)` will throw a `TypeError`.

As a rule of thumb, if you *ever* see `typeof blah === 'object'` in a codebase, it is probably wrong, unless accompanied by a guard (eg. `blah && typeof blah === 'object'`).

wincent: Sigh. If this comment doesn't appear to make any sense, it's because I didn't see that you already had `CKEDITOR.ADDITIONAL_RESOURCE_PARAMS &&` — I blame GitHub's very narrow UI.

julien: yes I remember we actually talked about this a few time. 
You can blame GitHub's UI ... or me for doing force push all the time

julien:
> If the default value is indeed always {}, I think that being defensive about non-object values is just noise.

Sure, I'll edit that

PR #91, opened 2020-07-08
docs: fix broken links in changelog

wincent:
Some of these were missing the necessary "v" prefix; that shouldn't happen in the future because of [this fix](https://github.com/liferay/liferay-npm-tools/pull/472).

In this commit, I'm just going back and fixing the broken links. Note that not all releases have the "v" prefix, because apparently we didn't prepare those releases in a consistent way:

- 4.13.1-liferay.4
- 4.13.1-liferay.1
- 4.11.4-liferay.1
- 4.11.3-liferay.0

We've had a ".yarnrc" with the "v" prefix set via "version-tag-prefix" since 1e1d26aeced (2020-02-13) though, so that should have at least prevented 4.13.1-liferay.4 from going out without a "v" prefix; I am guessing that that one was not published by the standard means (ie. by running `yarn version`), but we should make a point of always doing it that way (we've documented it in the README, which hopefully will be enough).


wincent:
cc @julien @carloslancha

PR #92, opened 2020-07-09
feat: hide button labels when editor size is small

mateomustapic:
The goal of this task is to hide button labels when editor size is small.
I made the changes by modifying `preset.css` files in all skins of `liferay-ckeditor`. 
This sheet contains custom styles specifically for these buttons
- https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/skins/kama/presets.css
- https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/skins/moono-lisa/presets.css
- https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/skins/moono/presets.css
- https://github.com/liferay/liferay-ckeditor/blob/master/skins/moono-lexicon/presets.css

As it is visible ine the linked files above, all these files contain duplicated styles, so I added styles following the same pattern.
Adding `768px` media query breakpoint with adding `display:none` on `cke_button_label`, `cke_button__source_label` and `cke_button__sourcedialog_label` classes does the trick.

### DIFF
--- a/skins/moono-lexicon/presets.css
+++ b/skins/moono-lexicon/presets.css
@@ -9,6 +9,15 @@ For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
 	display: inline;
 }
 
+/* Hide button labels on 768px breakpoint */
+@media (max-width: 768px) {
+	.cke_button__label,
+	.cke_button__source_label,
+	.cke_button__sourcedialog_label {
+		display: none;
+	}
+}
+
 /* "Font Size" panel size */
 .cke_combopanel__fontsize {
 	width: 135px;


julien:
@wincent I probably confused @mateomustapic this morning because I was explaining the "patch" workflow to him  as I thought he was going to modify the exiting `ckeditor-dev` skins too and not just our own `moono-lexicon` skin.

@mateomustapic I think it's OK to do this just in our `moono-lexicon` skin since that's what we'll end up using (at least in `liferay-portal`) - if that's the case, you should do what @wincent  suggested.

Regardless, I'm going to test this right now

patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
```diff
@@ -25,5 +25,5 @@ index 1d4e036fc..a53a4c809 100644
  				return;
  
 -- 
-2.27.0
+2.24.3 (Apple Git-128)
```
julien, line 28: > Apple Git-128 

Funny they didn't use `iGit`

wincent: And so necessary of them to call out that they've made their own little special build...

snowflake

julien:
@mateomustapic, this is working and looking good


Here's a screenshot of the editor in knowledge-base-web with the current skin 

ckeditor-knowledge-base-moono-lisa

We can see that the "Source" label is placed on the right of the source button

Here's a second screenshot, using a local build of `liferay-ckeditor` with these changes

ckeditor-knowledge-base-moono-lexicon

No more "Source" label and that's what we want. 

I just have a question: are we sure that only the source button in CKEditor is going to have a label?

(I checked and that's also the only one I see in the `ckeditor-dev` submodule)

I think we'll just need to pay attention if we add more plugins

julien:
I know @carloslancha is probably going to push changes related to the skin, so I'm just mentioning him here so he's aware about this.

mateomustapic:
> @mateomustapic, this is working and looking good
> 
> Here's a screenshot of the editor in knowledge-base-web with the current skin
> 
> ckeditor-knowledge-base-moono-lisa
> 
> We can see that the "Source" label is placed on the right of the source button
> 
> Here's a second screenshot, using a local build of `liferay-ckeditor` with these changes
> 
> ckeditor-knowledge-base-moono-lexicon
> 
> No more "Source" label and that's what we want.
> 
> I just have a question: are we sure that only the source button in CKEditor is going to have a label?
> 
> (I checked and that's also the only one I see in the `ckeditor-dev` submodule)
> 
> I think we'll just need to pay attention if we add more plugins

Hi Julien, thanks for checking this. 
`cke_button__label` is the mutual css class for all button labels in toolbar. 
`cke_button__source_label` is the class specific for source plugin. 
If we add and additional plugins have a label and different class then we would need to modify the styles in plugin's submodule.

julien:
I'm ok with this change - @carloslancha since you're also working on the skin I'd feel better if you could add your approval before merging. Thanks

julien:
@mateomustapic just don't forget to remove the changes in the `patches` directory

carloslancha:
@mateomustapic could you please add to the description the LPS/IFI/Whatever ticket is this about?

mateomustapic:
> @wincent I probably confused @mateomustapic this morning because I was explaining the "patch" workflow to him as I thought he was going to modify the exiting `ckeditor-dev` skins too and not just our own `moono-lexicon` skin.
> 
> @mateomustapic I think it's OK to do this just in our `moono-lexicon` skin since that's what we'll end up using (at least in `liferay-portal`) - if that's the case, you should do what @wincent suggested.
> 
> Regardless, I'm going to test this right now

I did make changes in `ckeditor-dev`, but for some reason the changes were not pushed in this PR.
I will modify the stuff again, patch and push the changes in few mins.

julien:
@mateomustapic I'm repeating myself, but I think it's OK to just do this just is the moono-lexicon skin for now
@carloslancha do you agree with that?

mateomustapic:
> @mateomustapic I'm repeating myself, but I think it's OK to just do this just is the moono-lexicon skin for now
> @carloslancha do you agree with that?

ok, then I'll do it just in moono-lexicon skin if Carlos agrees.

skins/moono-lexicon/presets.css
```diff
@@ -9,6 +9,15 @@ For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
 	display: inline;
 }
 
+/* Hide button labels on 768px breakpoint */
+@media (max-width: 768px) {
```
carloslancha, line 13: Is this what we want to do?

> the goal of this story is to limit the information that is shown in CKEditor's toolbar when the space allowed is not enough by hiding the buttons text.

With the media-query we're only looking for the window size, not the size of the editor. If the editor's container is smaller than 768 but the window is not this is result:

image

carloslancha:
Hey @mateomustapic 

> I did make changes in ckeditor-dev, but for some reason the changes were not pushed in this PR.

You're not supposed to push any changes made in `ckeditor-dev`, that's a submodule linked to `ckeditor` original repo, that's why we have the `patch` script. If you need to fix/change something of the original implementation of CKEeditor you:
- Change it in `ckeditor-dev`
- Run `sh ck.sh patch`
- Commit and push those patch files, not anything inside `ckeditor-dev`

Anyway, in this case, yep, you only need to change our `moono-lexicon` skin.

I've added a comment in the [JIRA ticket](https://issues.liferay.com/browse/LPS-116683) to be sure that this is the right way to implement this change.

skins/moono-lexicon/presets.css
wincent: this

mateomustapic:
> Hey @mateomustapic
> 
> > I did make changes in ckeditor-dev, but for some reason the changes were not pushed in this PR.
> 
> You're not supposed to push any changes made in `ckeditor-dev`, that's a submodule linked to `ckeditor` original repo, that's why we have the `patch` script. If you need to fix/change something of the original implementation of CKEeditor you:
> 
> * Change it in `ckeditor-dev`
> * Run `sh ck.sh patch`
> * Commit and push those patch files, not anything inside `ckeditor-dev`
> 
> Anyway, in this case, yep, you only need to change our `moono-lexicon` skin.
> 
> I've added a comment in the [JIRA ticket](https://issues.liferay.com/browse/LPS-116683) to be sure that this is the right way to implement this change.

Ok, thank you these instructions.

skins/moono-lexicon/presets.css
mateomustapic: > Is this what we want to do?
> 
> > the goal of this story is to limit the information that is shown in CKEditor's toolbar when the space allowed is not enough by hiding the buttons text.
> 
> With the media-query we're only looking for the window size, not the size of the editor. If the editor's container is smaller than 768 but the window is not this is the result:
> 
> image

The other way, of getting the editor container size instead of using media queries would me tricky. 
I think this solution would be sufficient in this case. 
This way we handled the responsive behaviour and resizing of editor respectfully in relation with window size and visibility of button labels which falls in the acceptance criteria.

jbalsas: Hey @mateomustapic, while this might be a good compromise, I agree with @carloslancha that this is not enough.

One possible solution to avoid media-queries and still keep a simple implementation would be:
- On resize, add custom breakpoint classes based on editor size (`editor-sm`, `editor-lg`, `editor-xl`)
- Use those classes instead of media queries to apply specific styles to breakpoint sizes

That still doesn't account exactly for "not having enough space" but I think would be more acceptable than relying on window size in general.

wincent: > The other way, of getting the editor container size instead of using media queries would me tricky and unnecessary.
> I think this solution would be sufficient in this case.
> This way we handled the responsive behaviour and resizing of editor respectfully in relation with window size and visibility of button labels which falls in the acceptance criteria.

I think you're misreading the acceptance criteria:

> As described in [Hiding Text](https://docs.google.com/document/d/1UPxckZRglY39i0UK6jEViv16C34HGRkTWxoCwqcr-kY), the goal of this story is to limit the information that is shown in CKEditor's toolbar when the space allowed is not enough by hiding the buttons text.
>
> Acceptance Criteria
> When the editor is resized and it starts wrapping
> Then Buttons no longer show labels inside them and only their icons remain

Let's not ship a half-baked fix that doesn't address the actual problem.

mateomustapic: > Hey @mateomustapic, while this might be a good compromise, I agree with @carloslancha that this is not enough.
> 
> One possible solution to avoid media-queries and still keep a simple implementation would be:
> 
> * On resize, add custom breakpoint classes based on editor size (`editor-sm`, `editor-lg`, `editor-xl`)
> * Use those classes instead of media queries to apply specific styles to breakpoint sizes
> 
> That still doesn't account exactly for "not having enough space" but I think would be more acceptable than relying on window size in general.

Ok, I will work on this suggested solution, thanks!

markocikos:
Followup in https://github.com/liferay-frontend/liferay-portal/pull/182

PR #95, opened 2020-07-10
fix: remove positioning and size definition of the editable area as is notlonger needed and was interfering with alloy editor styling

carloslancha:
We changed the skin Alloy Editor is using for CKEditor in portal to `moono-lisa` since its styling was made based on that one, but we're still having problems when both Alloy Editor and CKEditor are in the page 'cause they share classnames such as `ckeditor_editable`:

image

Removing this css properties, that are not longer needed, fixes this particular problem.


marcoscv-work:

marcoscv-work:

PR #96, opened 2020-07-16
fix: add border property to outer class required for resizing

mateomustapic:
This PR is the fix of the https://issues.liferay.com/browse/LPS-117074 bug when CKEditor breaks on resize. 
This bug occured because `moono-lexicon` skin’s `cke_chrome` class is missing border.
It is required because `resize` takes outer element’s border property to calculate width and height when resizing.

### DIFF
--- a/skins/moono-lexicon/mainui.css
+++ b/skins/moono-lexicon/mainui.css
@@ -48,6 +48,7 @@ Special outer level classes used in this file:
 /* The outer boundary of the interface. */
 .cke_chrome {
 	/* This is <span>, so transform it into a block.*/
+	border: 1px solid transparent;
 	display: block;
 	padding: 0;
 }


carloslancha:
LGTM.

Could you please add a link to the ckeditor code where resize calculations are being done? I'm curious about it :)

mateomustapic:
> LGTM.
> 
> Could you please add a link to the ckeditor code where resize calculations are being done? I'm curious about it :)

https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/core/tools.js#L899
If I am correct, this is the function where calculations are done while resizing.

carloslancha:
Merging this, feel free to release a new version if you need this to be fixed in portal asap. Let me know if you need any help ;)

PR #101, opened 2020-08-03
fix: remove comments from svg icons

carloslancha:
Solving https://issues.liferay.com/browse/LPS-118249

Remove comments from svg icons on css classes generation.

Copyright comments were introduced to Clay svg on https://github.com/liferay/clay/commit/13ae645d0fd90da6be1100f11430af259ff2f2c7#diff-214a1e2a1d78797d9528f65bfa7504ee

### DIFF
--- a/support/iconsClassesGenerator.js
+++ b/support/iconsClassesGenerator.js
@@ -24,6 +24,7 @@ const encodeSvgData = (svgData, color) => {
 	const symbols = /[\r\n%#()<>?\[\\\]^`{|}]/g;
 
 	return svgData
+		.replace(/<!--.*?-->/gs, '')
 		.replace(/(<svg[^>]*)( fill="[^"]+")/, '\1')
 		.replace(/<svg/, `<svg fill="${color}"`)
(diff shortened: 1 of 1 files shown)


jbalsas:
I see a regex... that can only mean one thing... we need @wincent!!

hackerman

support/iconsClassesGenerator.js
```diff
@@ -24,6 +24,7 @@ const encodeSvgData = (svgData, color) => {
 	const symbols = /[\r\n%#()<>?\[\\\]^`{|}]/g;
 
 	return svgData
+		.replace(/<!--.*-->/s, '')
```
wincent, line 27: You should use a non-greedy match here.

```
const text = '<!-- hi! -->contentz<!-- bye! -->';

const damaged = text.replace(/<!--.*-->/sg, ''); // ""

const cool = text.replace(/<!--.*?-->/sg, ''); // "contentz"
```

Also, I think you want a `g` to catch all comments.

carloslancha: Well, that was close 😅

PR #102, opened 2020-08-05
feat: add toolbarwatcher plugin

julien:
In order to solve [https://issues.liferay.com/browse/LPS-116683](https://issues.liferay.com/browse/LPS-116683),
and based on [https://github.com/liferay-frontend/liferay-portal/pull/182](https://github.com/liferay-frontend/liferay-portal/pull/182),
add a new plugin to control the visibility of toolbar labels in CKEditor

**Screenshot:**
Peek 2020-08-05 09-08


plugins/toolbarwatcher/plugin.js
```diff
@@ -0,0 +1,128 @@
...
+
+				if (toolbarsContainer) {
+					toolbarsContainerWidth =
+						parseInt(
+							getComputedStyle(toolbarsContainer).width,
+							10
+						) || 0;
+				}
+
+				return this.labeledToolbarsWidth >= toolbarsContainerWidth;
```
wincent, line 86: (Subjective, but) the `Infinity` here is a bit distracting because the thing reads as "magical special value" when I think the actual intent is "if I have a container, do a calculation with it, otherwise don't". So this might be clearer:

```js
if (toolbarsContainer) {
    var width = parseInt(
        getComputedStyle(toolbarsContainer).width,
        10
    ) || 0;

    return this.labeledToolbarsWidth >= width;
}

return false;
```

Up to you though.

plugins/toolbarwatcher/plugin.js
```diff
@@ -0,0 +1,128 @@
...
+				cancelDebounced(debounced);
+				debounced.id = setTimeout(function () {
+					fn.apply(null, args);
+				}, delay);
+			};
+		}
+
+		function cancelDebounced(debounced) {
+			clearTimeout(debounced.id);
+		}
```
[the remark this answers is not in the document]
julien: :+1:

plugins/toolbarwatcher/plugin.js
```diff
@@ -0,0 +1,128 @@
...
+			},
+
+			editor: null,
+
+			getToolbarsWidth() {
+				var containerElement = this.editor.container.$;
+
+				var toolbars = Array.from(
+					containerElement.querySelectorAll('.cke_toolbar')
+				);
```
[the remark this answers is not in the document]
julien: :+1:

ck.sh
```diff
@@ -56,6 +56,12 @@ case "$COMMAND" in
 				# Copy custom skin to ckeditor-dev
 				cp -r skins/moono-lexicon ckeditor-dev/skins/moono-lexicon
 
+				# Copy plugins
```
carloslancha, line 59: As we talk I think it would be better to replicate the structure in CKEditor

After this pr we'll have:

```
skins
   - moono-lexicon
      - plugins
         - balloonpanel
            - balloonpanel.css
         - balloontoolbar
            - balloontoolbar.css
         ...
plugins
   - toolbarwatcher
      - styles
         - toolbarwatcher.css
      - plugin.js
```

In this way we're mixing two concepts, forcing us to include this in the plugin
```
CKEDITOR.document.appendStyleSheet(
	this.path + 'styles/' + PLUGIN_NAME + '.css'
);
```
not being able to specify a style per skin for the plugin and having to add more build logic.

As I said IMHO we should replicate CKEditor structure:

```
- skins
   - moono-lexicon
      mainui.css
      [NO PLUGINS FOLDER HERE]
      ...
- plugins
   - balloonpanel
      - skins
         - moono-lexicon
            - balloonpanel.css
      - plugin.js
   - toolbarwatcher
      - skins
         - moono-lexicon
            - toolbarwatcher.css
      - plugin.js
```

We can go as it is right now and change it in other task if you want or do it in this one.

What do you guys think?

plugins/toolbarwatcher/plugin.js
julien: good point

ck.sh
julien: @carloslancha I agree that we could (and should) clean this up

But ... I don't agree 100% with this 

> In this way we're mixing two concepts, forcing us to include this in the plugin
> ```js
> CKEDITOR.document.appendStyleSheet(
> 	this.path + 'styles/' + PLUGIN_NAME + '.css'
> );

> not being able to specify a style per skin for the plugin and having to add more build logic.

We're not forcing anything here, I just thought that this plugin should have it's own styles defined and I also think that this behavior shouldn't depend on the skin, in fact as long as you enable this plugin, no matter what skin is being used the plugin should still behave the same.

And then I have a doubt about your second "file listing", i.e.

```
- skins
   - moono-lexicon
      mainui.css
      [NO PLUGINS FOLDER HERE]
      ...
- plugins
   - balloonpanel                                      <-    Why is balloonpanel here?*
      - skins                                             
         - moono-lexicon
            - balloonpanel.css
      - plugin.js
   - toolbarwatcher
      - skins
         - moono-lexicon
            - toolbarwatcher.css
      - plugin.js
```

The `balloonpanel` is not **ours** so I'm confused why you listed it here
This new `plugin` directory is for placing our own plugins.

Finally, as far as I'm concerned, I think we should do this in another PR

julien: @carloslancha one last question:

How do you plan to include plugin specific css in our `moono-lexicon` skin?

jbalsas: Not that I don't love this discussion... but for this CSS, plugins usually would simply do:

```javascript
onLoad: function() {
    CKEDITOR.addCss(`
        .cke_hide_toolbar_labels .cke_button__label,
        .cke_hide_toolbar_labels .cke_button__source_label,
        .cke_hide_toolbar_labels .cke_button__sourcedialog_label {
	    display: none;
        }`
    );
}
```

That seems to be the canon way to do this for simple CSS... and maybe pushes this conversation into the future 🤷

carloslancha: >The balloonpanel is not ours so I'm confused why you listed it here
This new plugin directory is for placing our own plugins.

It was an example following the original CKEditor structure. The point is that in the original structure you don't have plugins styles in `skins/yourskin` but in `plugins/yourplugin/skins/yourskin`. So, in this case we do have styling for `balloonpanel` in `moono-lexicon`, and that css is in `skins/moono-lexicon/plugins/balloonpanel`. What I'm suggesting is that css shouldn't be there (in fact that there should not exist any plugins folder inside `skins/yourskin`) but in `plugins/balloonpanel/skins/moono-lexicon/`, trying to mimic original structure.

> How do you plan to include plugin specific css in our moono-lexicon skin?

We're doing it already. The script we have for the plugins styles is copying those files from `skins/moono-lexicon/plugins/yourplugin` to `ckeditor-dev/plugins` before building it. What I'm suggesting is move all of that to our `plugins` folder and just copy it to `ckeditor-dev` before build in the same way.

>We're not forcing anything here, I just thought that this plugin should have it's own styles defined and I also think that this behavior shouldn't depend on the skin, in fact as long as you enable this plugin, no matter what skin is being used the plugin should still behave the same.

You're right, is no non-skin-dependant. Taking a look to other similar plugins like autocomplete the structure is:

```
plugins
   - autocomplete
      - skins
         - default.css
```

and it's being defined by:

```
CKEDITOR.document.appendStyleSheet( this.path + 'skins/default.css' );
```

In other plugins that are skin dependant (like balloonpanel) the structure is:

```
plugins
   - autocomplete
      - skins
         - moono-lisa
            balloonpanel.css
```

and it's beinf defined by:

```
CKEDITOR.document.appendStyleSheet( this.path + 'skins/' + CKEDITOR.skin.name + '/balloonpanel.css' );
```

So, summarizing:

- We can take care of the cleaning and move those `skins/.../plugins` to `plugins` and use the same structure in other pr
- In this one I'd try to mimic original structure by renaming `plugins/toolbarwatcher/styles/toolbarwatcher.css` to `plugins/toolbarwatcher/skins/default.css

carloslancha: > Not that I don't love this discussion... but for this amount of CSS, plugins usually would simply do:

If we have the opportunity in this pr with not so much effort (just renaming as I suggested in the end of my last comment) to have separate js and css... I'd go for it instead of adding css in the js file in that way...

jbalsas: > If we have the opportunity in this pr with not so much effort (just renaming as I suggested in the end of my last comment) to have separate js and css... I'd go for it instead of adding css in the js file in that way...

What do see as fundamentally wrong in adding the css that way? There's an API for it, more plugins seem to use that approach than the other and it avoids an additional request to an additional file...

I was not suggesting that as a way to get out of having this conversation. I truly believe loading an additional css file for 1 line sounds a bit overkill to me... but I'm fine with whatever call you make

julien: @jbalsas you're right - and I even thought of doing that to make things more simple, but with @carloslancha we had a slack call and both agree that we're going to go with two separate files, one for the JS and another for CSS - unless you really disagree ...





@carloslancha so now that we've talked about this on a call and that we both understand what we want, what's left to do for this PR is to rename the CSS file and directory to make them more similar to other plugins included with CKEditor, and then I'll open an issue and create a PR for the other "thing" that we talked about which is just moving files and making sure everything works as expected

wincent: > unless you really disagree

The fate of the cosmos may hinge on us making the correct decision here, so I will add my 2 cents:

> There's an API for it ... **more plugins seem to use that approach** than the other and **it avoids an additional request**

Those two bolded bits sound like pretty good reasons to go with the `CKEDITOR.addCss` approach; also it seems like the most minimal delta from what you have here and I like minimal things due to laziness and FOBS (Fear Of Breaking Shit).

carloslancha: I don't see anything as fundamentally wrong, I just don't like having css defined in the js file because maintainability and readability. I wouldn't like ending up having something like in `plugins/image2`.

```
CKEDITOR.addCss(
			'.cke_image_nocaption{' +
				// This is to remove unwanted space so resize
				// wrapper is displayed property.
				'line-height:0' +
			'}' +
			'.cke_editable.cke_image_sw, .cke_editable.cke_image_sw *{cursor:sw-resize !important}' +
			'.cke_editable.cke_image_se, .cke_editable.cke_image_se *{cursor:se-resize !important}' +
			'.cke_image_resizer{' +
				'display:none;' +
				'position:absolute;' +
				'width:10px;' +
				'height:10px;' +
				'bottom:-5px;' +
				'right:-5px;' +
				'background:#000;' +
				'outline:1px solid #fff;' +
				// Prevent drag handler from being misplaced (https://dev.ckeditor.com/ticket/11207).
				'line-height:0;' +
				'cursor:se-resize;' +
			'}' +
			'.cke_image_resizer_wrapper{' +
				'position:relative;' +
				'display:inline-block;' +
				'line-height:0;' +
			'}' +
			// Bottom-left corner style of the resizer.
			'.cke_image_resizer.cke_image_resizer_left{' +
				'right:auto;' +
				'left:-5px;' +
				'cursor:sw-resize;' +
			'}' +
			'.cke_widget_wrapper:hover .cke_image_resizer,' +
			'.cke_image_resizer.cke_image_resizing{' +
				'display:block' +
			'}' +
			// Hide resizer in read only mode (#2816).
			'.cke_editable[contenteditable="false"] .cke_image_resizer{' +
				'display:none;' +
			'}' +
			// Expand widget wrapper when linked inline image.
			'.cke_widget_wrapper>a{' +
				'display:inline-block' +
			'}' );
```

If you really think avoid loading this css could make a difference... I'm not against it.

jbalsas: > I wouldn't like ending up having something like in plugins/image2

That sounds reasonable. I'm sure that if this ever started to reach that scale, then the maintainability argument would gain weight. It's **not your current case**, though, so it feels a little like putting the band-aid before being wounded (not sure this translates properly to english 😂 ).

carloslancha: Anytime I see that meme I'll stop talking. Just saying 😄

wincent: > I don't see anything as fundamentally wrong, I just don't like having css defined in the js file because maintainability and readability.

Now put your React hat on and think about how JSX mixes "markup" with "logic" and the colocation works brilliantly.

Year 2010: Separation of Concerns! HTML + CSS + JS

Year 2015: HTML + CSS + JS is not Separation of Concerns, it's Separation of Technologies! The component should be the fundamental unit of abstraction.

Year 2020: Yo, the function is the fundamental unit of abstraction, bro. Just sayin'.

When JSX first came out people thought it was crazy and made fun of it. Fast forward a few years, and there are _literally countless_ [CSS-in-JS](https://www.google.com/search?q=css+in+js) implementations... (continuing the cosmological references...) _as countless as the stars in the sky_...

stars

And of course GraphQL came along which is all about colocating data fetching queries with supposedly visual components too.

In short, you should embrace this change, not fight it.

join-us

julien: > What do see as fundamentally wrong in adding the css that way? There's an API for it, more plugins seem to use that approach than the other and it avoids an additional request to an additional file...

Ok forget what I said. Let's embrace `CKEDITOR.addCss`

carloslancha:
LGTM

julien:
re-submitted in #114

Issue #16, opened 2019-04-10
Define workflow for patching CKEditor


wincent:
This issue is to provide a place for us to discuss how we want to approach the question of applying patches to the upstream CKEditor given our new build system. For context, the old and the new build system are described in https://github.com/liferay/liferay-ckeditor/issues/7.

# Ways to modify CKEditor

* **Submit a patch upstream:** This is the ideal scenario for bug fixes of a general nature, because it frees us from having to maintain a fork. On the flip side however, waiting for patches to be published in a released version adds a delay that we don't necessarily control, and not all patches may be accepted.
* **Use runtime configuration to modify behavior:** We can (and do) use plug-ins to extend or modify the behavior of the editor without modifying its source code.
* **Use runtime monkey-patching to modify behavior:** This is a more aggressive version of the above, and involves mutating or circumventing the normal functioning of the editor to produce a different result. This is sometimes a cheap and effective measure, but it can be quite difficult to pull off, may result in shipping useless code to the client, and can easily break as we move between versions.
* **Maintain a fork:** This is basically what we used to do in this repo: the commit graph of the upstream project is in the repo, and then we add commits on top of it; because it's a Git repo, we can use whatever internal structure (branches, tags, directories etc) or tools (cherry-picking, merging etc) that we see fit.
* **Maintain a patchset:** This is a variation of the fork approach. Instead of adding commits directly to the commit graph, we keep patches (for example, it could be patches created with `git-format-patch`) in a directory, and apply them as part of the build process. In this approach there are two main ways to get the base version of the code upon which to apply the patches: one is from a submodule (ie. what we currently have in this repo) and the other would be to keep it on a branch (ie. you'd add the CKEditor upstream as a remote and just create a local tracking branch to follow it here).

Observations:

* All these approaches have one thing in common: it is painful to maintain a long-lived set of modifications on top of a third-party project in motion.
* Because all the approaches suck in different ways, I don't think we should obsess over finding a perfect "silver bullet" solution, but rather, find an adequate solution and then provide clear documentation and workflows for applying it.

# Proposals

To get the ball rolling on the discussion here, I'll detail a proposal for what I think would be a decent solution.

# Proposal 1: Submodule plus explicit patch-list with supporting scripts

* Our default policy for actual bugs should be to try and get an upstream fix accepted whenever possible so that we don't have to maintain a long-lived patch that keeps us divergent from the upstream project.
[...]

Issue #7, opened 2019-03-20
Upgrade to latest CKEditor v4 (eg. ~4.11.3)


wincent:
I'm creating this issue to lay out a plan for moving us from the version of CKEditor that we are currently using in Liferay DXP (v4.6.0 + our patches) to the latest upstream v4 release (at the time of writing, that's v4.11.3). The corresponding internal ticket is: https://issues.liferay.com/browse/LPS-91153

# How things work right now

This repo is a fork of [the upstream ckeditor-dev repo](https://github.com/ckeditor/ckeditor-dev), based on [the v4.6.0 release](https://github.com/ckeditor/ckeditor-dev/tree/4.6.0), with a number of patches applied on top in the [ee-4.6.x branch](https://github.com/liferay/liferay-ckeditor/tree/ee-4.6.x). As an indication of the magnitude of the changes we've applied, this is the `git diff --stat` summary:

```
 33 files changed, 1210 insertions(+), 40 deletions(-)
```

Because this is a fork of the upstream repository, we can use [its `build.sh` script](https://github.com/liferay/liferay-ckeditor/blob/018f38915674303cb09347f7501b67e15d3befff/dev/builder/build.sh) in conjunction with [our config](https://github.com/liferay/liferay-ckeditor/blob/018f38915674303cb09347f7501b67e15d3befff/config.js) to produce a final build that bundles about 50 standard plug-ins. We do this with [a gulpfile](https://github.com/liferay/liferay-ckeditor/blob/018f38915674303cb09347f7501b67e15d3befff/gulpfile.js) that [specifies](https://github.com/liferay/liferay-ckeditor/blob/018f38915674303cb09347f7501b67e15d3befff/gulpfile.js#L2) an `artifactName` of "ckeditor" and offers these tasks:

```
[13:08:44] Tasks for ~/code/liferay-ckeditor/gulpfile.js
[13:08:44] ├── clean-maven-dist
[13:08:44] ├── prepare-maven-install
[13:08:44] ├── prepare-maven-snapshot
[13:08:44] ├── init-maven-install
[13:08:44] ├── install-maven-release
[13:08:44] ├── prepare-maven-artifact
[13:08:44] ├── publish-maven-artifact
[13:08:44] ├── publish-maven-snapshot
[13:08:44] ├── maven-install
[13:08:44] ├── maven-publish
[13:08:44] └── maven-publish-snapshot
``` 

[...]

PR #108, opened 2020-08-06
fix: Don't pass languageId to css files requests

carloslancha:
Fixing https://issues.liferay.com/browse/LPS-118624

Patching ckeditor so `languageId` is not passed to `css` requests

### DIFF
--- a/patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
+++ b/patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
@@ -1,0 +1,25 @@
+From a62acb27ffd532f3c3bb00e9ad727b62a95e9af6 Mon Sep 17 00:00:00 2001
+From: Carlos Lancha <carlos.lancha@liferay.com>
+Date: Thu, 6 Aug 2020 14:42:21 +0200
+Subject: [PATCH 6/6] LPS-118624 Don't pass languageId to css files requests
+
+---
+ core/ckeditor_base.js | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/core/ckeditor_base.js b/core/ckeditor_base.js
+index a6ce37594..6d9fcbc03 100644
+--- a/core/ckeditor_base.js
++++ b/core/ckeditor_base.js
+@@ -184,7 +184,7 @@ if ( !window.CKEDITOR ) {
+ 						var key = entries[0];
+ 						var value = entries[1];
+ 
+-						if (!resourceUrl.searchParams.has(key)) {
++						if (!resourceUrl.searchParams.has(key) && !(resourceUrl.toString().includes('.css') && key === 'languageId')) {
+ 							resourceUrl.searchParams.set(key, value);
+ 						}
+ 					});
+-- 
+2.23.0
+
--- a/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
+++ b/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
@@ -1,7 +1,7 @@
-From 8bb123947bf25d742f076fcd9facb94daa03fb27 Mon Sep 17 00:00:00 2001
+From bb8157502a44ede7c8b1806ef63fcdf3554e8089 Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Tue, 14 May 2019 10:47:25 +0200
-Subject: [PATCH 1/5] LPS-89596 Cannot Drag Image from top content line in IE11
+Subject: [PATCH 1/6] LPS-89596 Cannot Drag Image from top content line in IE11
  using Alloy Editor
 
 Changed position of drag icon for IE.
@@ -25,5 +25,5 @@ index 1d4e036fc..a53a4c809 100644
  				return;
  
 -- 
-2.27.0
+2.23.0
 
--- a/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
+++ b/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
@@ -1,7 +1,7 @@
-From 7f7f913b44304631f18fa2e9ecac7784adf77452 Mon Sep 17 00:00:00 2001
+From 30fce10a6a6b640ab317ad7cb96a540940afdcd2 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Roland=20P=C3=A1kai?= <roland.pakai@liferay.com>
 Date: Tue, 21 May 2019 09:38:15 +0200
-Subject: [PATCH 2/5] LPS-95472 Tabs in popups not appears correctly in
+Subject: [PATCH 2/6] LPS-95472 Tabs in popups not appears correctly in
  maximized CKEditor
 
 Use `width: inherit` instead of `width: 100%` to work with the maximize
@@ -24,5 +24,5 @@ index 022d7b473..5537eef39 100644
  				contentMap = this._.contents[ contents.id ] = {},
  				cursor,
 -- 
-2.27.0
+2.23.0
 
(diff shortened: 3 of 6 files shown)


patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/core/ckeditor_base.js b/core/ckeditor_base.js
+index a6ce37594..6d9fcbc03 100644
+--- a/core/ckeditor_base.js
++++ b/core/ckeditor_base.js
+@@ -184,7 +184,7 @@ if ( !window.CKEDITOR ) {
+ 						var key = entries[0];
+ 						var value = entries[1];
+ 
+-						if (!resourceUrl.searchParams.has(key)) {
++						if (!resourceUrl.searchParams.has(key) && !(resourceUrl.toString().includes('.css') && key === 'languageId')) {
```
julien, line 19: @carloslancha  if I understand this correctly, that means that only the `languageId` "extra" param can be passed.
Is that what we want ?

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
```diff
@@ -0,0 +1,25 @@
...
+
+diff --git a/core/ckeditor_base.js b/core/ckeditor_base.js
+index a6ce37594..6d9fcbc03 100644
+--- a/core/ckeditor_base.js
++++ b/core/ckeditor_base.js
+@@ -184,7 +184,7 @@ if ( !window.CKEDITOR ) {
+ 						var key = entries[0];
+ 						var value = entries[1];
+ 
+-						if (!resourceUrl.searchParams.has(key)) {
```
jbalsas, line 18: Patches are quite hard to review... but I'd say it'd be better abort before even starting the the loop sequence

```
Object.entries(CKEDITOR.ADDITIONAL_RESOURCE_PARAMS).forEach(function(entries) {
    ...
}
```

No reason to check the same for each key.

wincent: > Patches are quite hard to review

Pro-Tip™: click the "View file" link (in the ... menu for the patch) to see the change without all the noise: https://github.com/liferay/liferay-ckeditor/blob/80e862ef5f0704180006404d229151f047e66688/patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch

jbalsas: Yeah, what I meant is you don't have the interactive diff, so if you want to see where is that `if` you have to check the source file, right? 

And you don't know what other patches are applied on top, so you need to check the source file plus there result of the previous patches to know for sure how does the surrounding code look like?

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
julien: @carloslancha forget what I just said, I didn't read that correctly

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
carloslancha: M... but we want to not add the param only if the key is `languageId`. If we abort it before iterating we won't add any param

jbalsas: Yeah, that's a fair point, I was thinking just skip all params for css files, but this should do the trick too

wincent:
> I see changes in different `.patch` files, not sure if that's expected or not. If it is, it feels odd. I get it's auto-generated and seems to depend on the overall collection of patches... but it does feel like it would make adding a new one harder over time.

Yep, it's expected. Some of it is just noise (eg. user doing the update is using a different version of Git, so the version number changes), and some of it is due to what you said (eg. total number of patches changes, so each subject must change too). We can reduce the version-number-related noise by passing `--no-signature` to our `git format-patch` call, and we can reduce the noise from the numbering by passing `--no-numbered` to our `git format-patch` call.

Note though that we can't really stop the SHA-1's from changing if a rebase or other mutation is involved, but they will probable stay stable otherwise. Let me send a quick PR to test this out.

carloslancha:
> On a different note... I see changes in different .patch files, not sure if that's expected or not. If it is, it feels odd. I get it's auto-generated and seems to depend on the overall collection of patches... but it does feel like it would make adding a new one harder over time.

I just followed the patching instructions and taking a look to previous patches commits [like this one](https://github.com/liferay/liferay-ckeditor/commit/7ee7fb1c5a0344fda6174ac79bdfa5a3e462df7b) looks like it's correct.

@julien could you please confirm?

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
jbalsas: It's a bit weirder because it's harder to explain that `if` now... people coming in the future (us) won't understand why simply by the message `fix: Don't pass languageId to css files requests`. We will understand what, but not what was so special about css files and the `languageId` param.

carloslancha:
Ok, @wincent commented just 19 seconds before sending my comment xD

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
carloslancha: 👍

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
carloslancha: Roger that

jbalsas:
toomuchpeople

wincent:
Cleaner patches: https://github.com/liferay/liferay-ckeditor/pull/109

Demo: https://github.com/liferay/liferay-ckeditor/commit/0b67c64af91dd710177fcf8372e02a62f9d9f7d5

carloslancha:
@jbalsas done

jbalsas:
makeitso

jbalsas:
wait

jbalsas:
@wincent is pushing DO NOT MERGE 😂

wincent:
> @wincent is pushing DO NOT MERGE 😂

Eh?

wut

I just said don't merge [this commit](https://github.com/liferay/liferay-ckeditor/commit/0b67c64af91dd710177fcf8372e02a62f9d9f7d5), which is a demo.

After this is in, merge #109 and regenerate to get cleaner patches.

jbalsas:
> Eh?
>
> wut

looks_wrong

I got confused by the UI

Screen Shot 2020-08-06 at 16 09 00

Thought you'd pushed to this branch, not to another branch a commit that had a reference to this one...

patches/0006-LPS-118624-Don-t-pass-languageId-to-css-files-reques.patch
wincent: > Yeah, what I meant is you don't have the interactive diff, so if you want to see where is that `if` you have to check the source file, right?
> 
> And you don't know what other patches are applied on top, so you need to check the source file plus there result of the previous patches to know for sure how does the surrounding code look like?

If it's a non-trivial patch, your best bet is probably to pull it down the PR locally and run `./ck.sh setup`. In the ckeditor-dev submodule you can then see the commits that the patches correspond to and explore the file-system etc. If it is non-trivial too, you may want to build and test anyway, so pulling it down might be something you want to do anyway.

There's a definite trade-off here. We explored a bunch of options when we switched to the patch-based workflow, which you can read about [here](https://github.com/liferay/liferay-ckeditor/issues/16). We can obviously try one of the other options on that list (or perhaps something not even on it), if this one is too painful.

jbalsas: nah, it's okay, I was just passing by

wincent: If you want to know real pain, read about [how it used to be](https://github.com/liferay/liferay-ckeditor/issues/7).

pain

jbalsas: lifeispain

julien:
@carloslancha just wanted to confirm your patch is applied correctly :+1:

PR #114, opened 2020-08-07
feat: add toolbarwatcher plugin

julien:
In order to solve https://issues.liferay.com/browse/LPS-116683,
and based on liferay-frontend/liferay-portal#182,
add a new plugin to control the visibility of toolbar labels in CKEditor

**Screenshot**

toolbarwatcher

### DIFF
--- a/plugins/toolbarwatcher/plugin.js
+++ b/plugins/toolbarwatcher/plugin.js
@@ -1,0 +1,124 @@
+/**
+ * Copyright (c) 2000-present Liferay, Inc. All rights reserved.
+ *
+ * This library is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation; either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * This library is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ */
+
+(function () {
+	'use strict';
+
+	var PLUGIN_NAME = 'toolbarwatcher';
+
+	if (!CKEDITOR.plugins.get(PLUGIN_NAME)) {
+		var TOOLBAR_HIDDEN_LABELS_CLASS = 'cke_hide_toolbar_labels';
+
+		function debounce(fn, delay) {
+			return function debounced() {
+				var args = Array.prototype.slice.call(arguments);
+				clearTimeout(debounced.id);
+				debounced.id = setTimeout(function () {
+					fn.apply(null, args);
+				}, delay);
+			};
+		}
+
+		CKEDITOR.plugins.add(PLUGIN_NAME, {
+			afterInit: function (editor) {
+				this.editor = editor;
+				this.editor.on(
+					'instanceReady',
+					this.onInstanceReady.bind(this)
+				);
+			},
+
+			editor: null,
+
+			getToolbarsWidth: function () {
+				var containerElement = this.editor.container.$;
+
+				var width = 0;
+
+				containerElement
+					.querySelectorAll('.cke_toolbar')
+					.forEach(function (element) {
+						width += element.offsetWidth;
+					});
+
+				return width;
+			},
+
+			isToolbarOverflowing: function () {
+				var toolbarsContainer = this.editor.container.$.querySelector(
+					'.cke_top'
+				);
+
+				if (toolbarsContainer) {
+					var toolbarsContainerWidth =
+						parseInt(
+							getComputedStyle(toolbarsContainer).width,
+							10
+						) || 0;
+
+					return this.labeledToolbarsWidth >= toolbarsContainerWidth;
+				}
+
+				return false;
+			},
+
+			labeledToolbarsWidth: 0,
+
+			onInstanceReady: function (event) {
+				var instance = this;
+
+				this.labeledToolbarsWidth = this.getToolbarsWidth();
+
+				event.editor.window.on(
+					'resize',
+					debounce(function () {
+						var containerElement = instance.editor.container.$;
+						var isToolbarOverflowing = instance.isToolbarOverflowing();
+
+						if (isToolbarOverflowing) {
+							if (
+								!containerElement.classList.contains(
+									TOOLBAR_HIDDEN_LABELS_CLASS
+								)
+							) {
+								containerElement.classList.add(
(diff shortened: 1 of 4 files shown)


carloslancha:

carloslancha:
Hey @julien I've tested it in Blogs and Wiki and here you have two recordings about it:

In blogs it looks like resizing calculations are not working as expected and the change between toolbar sets based on screen size looks like is not working well either.
blogsresize

In wiki it looks like calculations are working as expected (with minor problems during resize due to the a11yHelpBtn), and same problems described above with toolbar sets are happening
wikiresize

Don't know if the toolbar sets stuff are related but I think it worth taking a look. Thx!

markocikos:
Why are we adding this plugin in `liferay-ckeditor` repo, an not in [portal `_diffs/plugins` dir](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins), like all other custom CKEditor plugins? Is this something we want start doing in the future? If so, we should move all other plugins for consistency.

julien:
@carloslancha thanks for the feedback: the calculations where based on [https://github.com/liferay-frontend/liferay-portal/pull/182](https://github.com/liferay-frontend/liferay-portal/pull/182) but I'll look into it.

@markocikos you're right: we should take a decision: with @carloslancha we were actually thinking that all our "custom" plugins should probably live in this (`liferay-ckeditor`) repository, but I'm opening the debate.

julien:
@carloslancha after talking with @markocikos we both think having our plugins in `liferay-portal` makes the developer experience easier - so we think this should go there as well.

carloslancha:
There was actually [a beginning of a discussion](https://github.com/liferay-frontend/liferay-portal/pull/182#issuecomment-667873404) about if we should move all `liferay-ckeditor` inside portal or not.

We all know the pros and cons about having a separated repo and about having it living in portal. In any case I'd go with having everything in the same place. It would be kinda confusing playing in here with the skins and with some plugins styles but having the actual plugins js code and other styles in portal...

If we want to move everything to `liferay-portal` we will need to create some infrastructure, so for now I'd stick to what we have now and try to place everything we can here.

wincent:
> we were actually thinking that all our "custom" plugins should probably live in this (liferay-ckeditor) repository

> we both think having our plugins in liferay-portal makes the developer experience easier

> I'd stick to what we have now and try to place everything we can here.

ping-pong

julien:
@carloslancha I removed the `debounce` in the `resize` event handler and it looks a bit better, that said before push any changes I'd like to understand what you mean with
>  the change between toolbar sets based on screen size looks like is not working well either.
because nothing has been modified concerning that

plugins/toolbarwatcher/plugin.js
```diff
@@ -0,0 +1,129 @@
...
+					});
+
+				return width;
+			},
+
+			init: function(editor) {
+				if (!stylesLoaded) {
+					CKEDITOR.document.appendStyleSheet(
+						this.path + 'skins/default.css'
+					);
```
wincent, line 64: I thought [you said](https://github.com/liferay/liferay-ckeditor/pull/102#discussion_r465661178):

> Let's embrace `CKEDITOR.addCss`

julien: I did, the problem is that sometimes these classes are added to the `iframe` and aren't applied, at least that's what @carloslancha told me

carloslancha: I tested this pr before that change and css was not being loaded, online inside the content iframe (where you actually write). Looks like extracting this to a css file did the trick, don't know if @julien found out why 😅

jbalsas:
Hey all!

> Why are we adding this plugin in liferay-ckeditor repo, an not in portal _diffs/plugins dir, like all other custom CKEditor plugins? Is this something we want start doing in the future? If so, we should move all other plugins for consistency.

I'd say we should move **ALL** plugins here, or wherever we develop CKEditor (right now, here). That allows us to properly integrate it in the build process (deciding which plugins to bundle by default and which ones to leave as external) and test in isolation of DXP (we should have a better workflow for this).

That has 1 caveat, though, which is plugins using DXP-specific APIs like the recent [addimages plugin](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins/addimages/plugin.js). The expectation for those plugins would be to expose some type of configurable callback to isolate generic CKEditor code from specific DXP behaviour.

See for example [how we change the default `documentBrowseLinkCallback` behaviour](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-alloyeditor-web/src/main/resources/META-INF/resources/alloyeditor.jsp#L147-L169).

carloslancha:
@jbalsas out! Enjoy your vacations!

plugins/toolbarwatcher/plugin.js
julien: @carloslancha the difference is that with `CKEDITOR.addCss` as you said the styles are only applied to CKEDITOR's iframe, with `CKEDITOR.document.appendStyleSheet` they're in the "top" document, i.e.

image

In blogs it doesn't seem to matter but I wanted to confirm that if we don't do this in wiki or knowledge base it breaks

carloslancha:

carloslancha:
Hey @julien 

I tested it and I found some issues:

If you resize the page to the point where labels are hide and you reload the page then the labels are shown, so maybe there's something wrong on running the watcher on instance ready. Image from blogs:

image

This is also happening when you resize the editor to the size where the toolbar set changes, together with some flickering (which is not appreciated in the following gif)

uIrip0187L

julien:
Hey @carloslancha I'm looking at this again, while waiting for #120 to merge.

I'm now testing in Web Content and Wiki to avoid testing where CKEditor isn't available by default.

I think it's pretty tricky to get this working exactly as we want to without having some "sizes" defined (i.e. one for desktop, one for tablet and one for mobile), but I don't see more "flickering" than before with this change

jbalsas:
Hey @julien, let's pause on this until after the release settles. We'll start working on this again in a few weeks.

julien:
@jbalsas ok, thanks for the heads up.
I added an "on-hold" label just to make sure nothing dangerous happens.

julien:
Feel free to close this 😘

bryceosterhaus:
@julien we will only close this if you come back to us

julien:
@bryceosterhaus 😂  I miss all of you but I don't miss ckeditor yet

izaera:
:cry:

PR #116, opened 2020-08-12
fix: fix img padding issue

mateomustapic:
This PR fixes https://issues.liferay.com/browse/LPS-80871

Img element in `.cke_editable` area did not have any padding, which this area ugly when `text` or another `img` is added.
Below is the comparison before/after this change.


mateomustapic:
before

mateomustapic:
after

julien:
@mateomustapic: 

**Concerning the pull request itself**

1. The commit title and message shouldn't use abbreviations:
    I know "img" means image but it would be better to use the word **image** to make things clearer

2. When you look at the commit message you'll see:

    > fix: fix img padding issue

    It's a small detail but the fact that the word *fix* is repeated looks weird to me, I'd rather see something like this

    > fix: add padding to images

3. The commit message should be changed, especially this line
    
    > Img element in .cke_editable area did not have any padding, which this area ugly when text or another img is added.

   > Img element

   Should probably be "Images"  

    > ".cke_editable area" 

    Should probably be "CKEditor" or "editor"

    > "which this area ugly" 

    I personally wouldn't use "ugly"

    > To test this
    > 
    > Wiki 
    > Add Wiki
    > Add Page
    > Edit added page
    
    I know more or less what that means, but do you think anyone could understand this? 

    I also think that as long as the page you're viewing is using `ckeditor` it should work: in Wiki but also Blogs or Web Content


**Concerning the actual change**

After building `liferay-ckeditor` and deploying that new version several times (just to make sure), I'm unable to see it working.

Here's what I see in Wiki for example:

after-wiki-01

And here's what I see in blogs (I've modified my `portal-ext.properties` file to use `ckeditor` everywhere):

after-blogs-01

Since you requested a review from @carloslancha as well, let's wait for his confirmation.

The LPS ([https://issues.liferay.com/browse/LPS-80871](https://issues.liferay.com/browse/LPS-80871)) mentions [this](https://issues.liferay.com/browse/LPS-78690) other issue, which has a reference to [this](https://github.com/brianchandotcom/liferay-portal/pull/59271) pull request: 
If you look at the [changes](https://github.com/brianchandotcom/liferay-portal/pull/59271/files), you'll notice that `margin` had been used to fix this (which makes sense) so why are we using `padding`?

I know that @carloslancha has worked on the skin, which means he probably knows the editor's CSS structure and will probably be glad to help if you have a problem with the styles not being applied correctly

carloslancha:
Hey guys! Not tested it, just some questions

- Was this working before `moono-lexicon` skin was applied?
- If so, what was different in `moono-lisa` (or in `liferay-portal`)? (a.k.a. what we broke? xD)

About the pr itself just one nit: It's preferable to place the images showing the before/after in the description instead of in different comments.

julien:
Hey @mateomustapic 

Actually the editor's styles were cached and that's why I wasn't seeing the change, so this is an updated screenshot: 

image

mateomustapic:
> @mateomustapic:
> 
> **Concerning the pull request itself**
> 
> 1. The commit title and message shouldn't use abbreviations:
>    I know "img" means image but it would be better to use the word **image** to make things clearer
> 2. When you look at the commit message you'll see:
>    > fix: fix img padding issue
>    
>    
>    It's a small detail but the fact that the word _fix_ is repeated looks weird to me, I'd rather see something like this
>    > fix: add padding to images
> 3. The commit message should be changed, especially this line
>    > Img element in .cke_editable area did not have any padding, which this area ugly when text or another img is added.
>    
>    
>    > Img element
>    
>    
>    Should probably be "Images"
>    > ".cke_editable area"
>    
>    
>    Should probably be "CKEditor" or "editor"
>    > "which this area ugly"
>    
>    
>    I personally wouldn't use "ugly"
>    > To test this
>    > Wiki
>    > Add Wiki
>    > Add Page
>    > Edit added page
>    
>    
>    I know more or less what that means, but do you think anyone could understand this?
>    I also think that as long as the page you're viewing is using `ckeditor` it should work: in Wiki but also Blogs or Web Content
> 
> **Concerning the actual change**
> 
> After building `liferay-ckeditor` and deploying that new version several times (just to make sure), I'm unable to see it working.
> 
> Here's what I see in Wiki for example:
> 
> after-wiki-01
> 
> And here's what I see in blogs (I've modified my `portal-ext.properties` file to use `ckeditor` everywhere):
> 
> after-blogs-01
> 
> Since you requested a review from @carloslancha as well, let's wait for his confirmation.
> 
> The LPS (https://issues.liferay.com/browse/LPS-80871) mentions [this](https://issues.liferay.com/browse/LPS-78690) other issue, which has a reference to [this](https://github.com/brianchandotcom/liferay-portal/pull/59271) pull request:
> If you look at the [changes](https://github.com/brianchandotcom/liferay-portal/pull/59271/files), you'll notice that `margin` had been used to fix this (which makes sense) so why are we using `padding`?
> 
> I know that @carloslancha has worked on the skin, which means he probably knows the editor's CSS structure and will probably be glad to help if you have a problem with the styles not being applied correctly

Hey @julien 
Thanks for the review.

Regarding the PR description and the commit message. 
> 1. The commit title and message shouldn't use abbreviations:

> 2. When you look at the commit message you'll see:
 > fix: fix img padding issue
To be honest I saw a PR from someone before with the same repeated detail, but of course I will fix it in the next push.

> 3. The commit message should be changed..
Your whole comment on this makes sense and I will follow the things you wrote.

Regarding the change 
I wrote steps to reproduce this in the Wiki, where I tested this change initially and later in Blogs and Web Content.
I didn't have a caching issue so I was able to see the change.
I used padding because I thought it would be better because image is alongside text in the same paragraph and since the `margin` is used to clear an area around an element (outside the border), and the `padding` clears an area around the content (inside the border) of an element, that's why I went with `padding`.

julien:
@mateomustapic 

> Regarding the change
I wrote steps to reproduce this in the Wiki, where I tested this change initially and later in Blogs and Web Content.

Yes, that's also how I understood it.
My point is that it's not easy to understand for someone else. Does that make sense? 

> I used padding because I thought it would be better because image is alongside text in the same paragraph and since the margin is used to clear an area around an element (outside the border), and the padding clears an area around the content (inside the border) of an element, that's why I went with padding.

So will the image still look "OK" if you add a border around it? 
My point was that if we had solved this before with `margin`;  I think we could probably apply the same fix.

mateomustapic:
@julien 
As requested, I used `margin` instead of `padding` and pushed the change (the border would be an issue if added).
Also, I have modified a commit message and a PR title as suggested.

skins/moono-lexicon/mainui.css
```diff
@@ -202,3 +202,7 @@ Special outer level classes used in this file:
 .cke_editable:focus {
 	background-color: #f0f5ff; /* $primary-l3 */
 }
+
+.cke_editable img {
+	margin: 24px;
```
carloslancha, line 207: Hey @mateomustapic did you see https://github.com/liferay/liferay-ckeditor/pull/116#issuecomment-673318822 ?

Other than that, I feel that this is not the solution we're looking for. Adding `margin: 24px` we're adding a margin in all the sides of the image but I think that what we want is to only add a margin to the left or to the right side depending on the selected alignment. That's why I was asking what was the previous behavior.

About the pr and the commit messages they say "add padding to images" when we're actually adding a margin ;)

mateomustapic: hi @carloslancha 
I missed the comment you mentioned, sorry, I will now take a look at the behaviour before applying `moono-lexicon` skin.

Regarding the commit message, I will fix it, rushing mistake :)

mateomustapic:
hey @julien 
After deploying old skin and comparing it with `moono-lexicon` I found out that this issue was present then and wasn't caused by skin changes.
To fix this issue correctly I think the approach from the mentioned PR https://github.com/brianchandotcom/liferay-portal/pull/59271 where the margin is added accordingly to image position is the right one.
I have tested this and if you agree this would be the right solution, please confirm so I can implement it.

carloslancha:
@mateomustapic go for it ;)

jbalsas:
Hey @julien, @mateomustapic, @carloslancha, I think we've been over this already in the past, but I'll mention it once again.

Visual styles applied inside the editor do not apply to content when presented outside the editor. Adding a padding in a CSS that only applies to the editor window would likely make the content look good while editing and bad when seen outside unless the same CSS is applied outside of CKEditor too. This is **not** a queue to also add the css fix to the outside world but to likely refrain from fixing it this way here.

mateomustapic:
Hey @jbalsas ,

I have been working on solution which doesn't include modifying the skin in `liferay-ckeditor`.
The main idea is to add
`.html-editor.portlet.cke_editable img {
	margin: 12px;
}`
in `frontend-theme-styled` theme, file _editor_html.scss => https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/frontend-theme/frontend-theme-styled/src/main/resources/META-INF/resources/_styled/css/application/_editor_html.scss
 
This solution is okay while editing content in the editor 
before

jbalsas:
Hey @mateomustapic!

> I have been working on solution which doesn't include modifying the skin in liferay-ckeditor.

The problem is not in modifying the skin, but that this change is **only** present within the CKEditor window but not outside. What happens when you click on publish? What's the margin/padding that the image has when that wiki page is being displayed and not edited? Try similarly with other apps like web content or knowledge base to see what I mean.

Visual changes need to be persisted in the document and not be a css modification because css may or may not be there at the time you present the stored HTML content. For example, this wouldn't even work inside CKEditor if someone was using a custom theme that didn't extend `styled`.

Please, let me know if this is still unclear!

mateomustapic:
> Hey @mateomustapic!
> 
> > I have been working on solution which doesn't include modifying the skin in liferay-ckeditor.
> 
> The problem is not in modifying the skin, but that this change is **only present within the CKEditor window but not outside**. What happens when you click on publish? What's the margin/padding that the image has when that wiki page is being displayed and not edited? Try similarly with other apps like web content or knowledge base to see what I mean.
> 
> Visual changes need to be persisted in the document and not be a css modification because css may or may not be there at the time you present the stored HTML content. For example, this wouldn't even work inside CKEditor if someone was using a custom theme that didn't extend `styled`.
> 
> Please, let me know if this is still unclear!

I see, it is clear now. You are right, the published content in this and similar editors wouldn't have this margin. 
I will work on solving this within the content document without modifying css. 
So, we need to persist this somehow so that when the user views the saved content he'll see it with the correct styles.
I am not sure how to do it yet, but I will focus on it. Marcos has suggested adding inline `style=`, but he would prefer leaving default margin and modifying the theme in this case.

marcoscv-work:
Do we need a margin by default here? I have the feeling we are trying to cover all possibilities needs the customer could have
I mean, we are trying to add a general margin to add space with texts but it is creating extra margins on the opposite side or after/before and probably I don't need them. What if what I want is to add three glued images?

It should be great if we can decide if we add a margin or not by adding a general helper class like `m-5` or `ml-5` (like content page editor) and then the customer changes those values by the theme or by custom properties.
if not, easier if we can dispense with the default margin and that the customers use their theme to decide the margin they want and where

jbalsas:
> Do we need a margin by default here? 

Definitely not. This should be something that can be configured per-image with a sensible default. I'd expect the {{image}} plugin in CKEditor to offer something like this: configurable padding or margin with a default. That's what we should be focusing on, imho.

mateomustapic:
> > Do we need a margin by default here?
> 
> Definitely not. This should be something that can be configured per-image with a sensible default. I'd expect the {{image}} plugin in CKEditor to offer something like this: configurable padding or margin with a default. That's what we should be focusing on, imho.

Hi @jbalsas ,

So, after researching on solving this issue without modifying the styles I found out the following:

Since we enabled the `image2` plugin in https://github.com/liferay-frontend/liferay-portal/pull/258
we lost the original image plugin's ability to add space around images.

Unfortunately we cannot use both `image` and `image2` plugins together, as you'll see here:
https://github.com/ckeditor/ckeditor4/blob/965f621e6ab15f7f62345b9b21785d9fdf02895d/plugins/image/plugin.js#L24-L26

I know that we want to avoid patching plugins as you mentioned here:
https://github.com/liferay/liferay-ckeditor/pull/117#issuecomment-673419276

But in this case, if we want to be able to center and resize images as well, which we wanted when enabling `image2`, along with adding space around them, we should do it by patching the plugin.

I also checked the 3rd image plugin, `easyimage`: https://ckeditor.com/cke4/addon/easyimage but it doesn't have all the requirements needed.

jbalsas:
Hey @mateomustapic, that sounds like a fair analysis of the situation. 

The `image` plugin allows configuring `HSpace` and `VSpace` and sets margins according to the values in those fields and the `image2` plugin lacks that.

Before going the patch way, there's an official API to affect dialogs which we might be able to use to enhance the image properties plugin. Please check for the `dialogDefinition` and see how we're using it in other places in DXP and try to figure out if that way you can achieve a similar effect.

PR #117, opened 2020-08-13
fix: add the ability to center images

julien:
This fixes [https://issues.liferay.com/browse/LPS-115747](https://issues.liferay.com/browse/LPS-115747).

The `image` plugin's dialog has been patched and now shows an option to "center" images

**Before**

before

**After**

centered-image

Note: you'll have to trust me that the `console.log` message that appear in the second .gif have been removed (although that's visible in the patch)

### DIFF
--- a/patches/0007-LPS-115747-Add-the-ability-to-center-images.patch
+++ b/patches/0007-LPS-115747-Add-the-ability-to-center-images.patch
@@ -1,0 +1,88 @@
+From 8e342f43622a2519a89304f19a44f2a399c8a8d5 Mon Sep 17 00:00:00 2001
+From: Julien Castelain <julien.castelain@liferay.com>
+Date: Thu, 13 Aug 2020 12:34:29 +0200
+Subject: [PATCH] LPS-115747 Add the ability to center images
+
+---
+ plugins/image/dialogs/image.js | 35 +++++++++++++++++++++++++++-------
+ 1 file changed, 28 insertions(+), 7 deletions(-)
+
+diff --git a/plugins/image/dialogs/image.js b/plugins/image/dialogs/image.js
+index acd7b66db..a9447b90e 100644
+--- a/plugins/image/dialogs/image.js
++++ b/plugins/image/dialogs/image.js
+@@ -911,7 +911,7 @@
+ 								},
+ 								{
+ 									id: 'cmbAlign',
+-									requiredContent: 'img{float}',
++									requiredContent: 'img{display,float,margin}',
+ 									type: 'select',
+ 									widths: [ '35%', '65%' ],
+ 									style: 'width:90px',
+@@ -920,6 +920,7 @@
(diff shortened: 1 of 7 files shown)


jbalsas:
please, check the description in the issues... I think I added detailed information about the different available plugins.

Let's switch to image2 and avoid patching image of possible

patches/0007-LPS-115747-Add-the-ability-to-center-images.patch
```diff
@@ -0,0 +1,88 @@
...
+ 												// Ignore those unrelated values.
+ 												case 'inherit':
+ 												case 'none':
+-													value = '';
++													floatValue = '';
++											}
++
++											var value = floatValue;
++
++											if (displayValue === 'block' && marginValue === '0px auto') {
```
carloslancha, line 51: When you use `0` no unit should be specified, so this should be `marginValue === '0 auto'`

patches/0007-LPS-115747-Add-the-ability-to-center-images.patch
```diff
@@ -0,0 +1,88 @@
...
+-											if ( value )
+-												element.setStyle( 'float', value );
+-											else
++											if ( value ) {
++												if ( value === 'left' || value === 'right' ) {
++													element.setStyle( 'float', value );
++												}
++												else if (value === 'center') {
++													element.setStyle('display', 'block');
++													element.setStyle('margin', '0px auto');
```
carloslancha, line 69: When you use `0` no unit should be specified, so this should be `'margin', '0 auto'`

julien:
@jbalsas I did read it, I think that concerning the different plugins they are described in https://issues.liferay.com/browse/LPS-115746 and not https://issues.liferay.com/browse/LPS-115747

So OK, I'll forget about patching and switch to image2

patches/0007-LPS-115747-Add-the-ability-to-center-images.patch
julien: Yeah I know, but CKEDITOR automatcially adds "px" @carloslancha ...

patches/0007-LPS-115747-Add-the-ability-to-center-images.patch
julien: @carloslancha given @jbalsas 's comment let's close this one. But thanks as always

julien:
Closing because we don't want to patch - since image2 solves this and will be easier than patching

jbalsas:
All the tasks in this epic are not about blindly implementing the requests rather than figuring out how to best leverage CKEditor's features.

For example, here it's not only about centering images in the dialog, but also in the center align toolbar button and command. 

Please, treat all these tasks as opportunities to investigate what CKEditor has to offer and work towards a coherent and unified UX that's also more maintainable for us in the long run.

PR #120, opened 2020-08-21
feat: add code mirror plugin

julien:
*This is still working in progress and I'm just submitting to gather feedback*

In order to complete [LPS-118290](https://issues.liferay.com/browse/LPS-118290), we're adding a CKEditor plugin named `codemirror` to mimic `alloyeditor`'s source view.

Instead of focusing only on the code, I thought it might be more easy to add some screenshots, in the context of `liferay-portal` and to do the actual code review in a separate PR.

**This screenshot shows the `codemirror` plugin working in CKEditor**

(*I know I made typo in that .gif but that's not the point for now*)

**As a reminder, CKEditor's original `sourcearea` plugin looks like this**

**Here's my first doubt**: Should `CodeMirror` be configured to disable attributes or non allowed tags ...
here's another .gif that might make it easier to understand:

**Now for the second part**: 

Since this plugin needs to behave like `alloyeditor`'s source view, we should be able to open a dialog when in source mode, a button will be available in CKEditor's toolbar but will only be enabled when in source mode. 

I didn't know what icon to choose for the button so I went ahead with the **preview** button, but I doubt we want to use it; here's a screenshot:

(*Note that the button itself does not have a red border around it, I added a red border when taking the screenshot to make sure everyone would see which button/icon I was referring to*)

Another issue I'm facing is that when opening the dialog, I'm getting an error in [dialog_definition.js](https://github.com/liferay-frontend/liferay-portal/blob/244f8cbbeb63f833306b14c2f34ad52929c6703d/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/extension/dialog_definition.js#L17), because `event.editor.window` is `undefined` ... it might be easier to understand with this .gif

I think there's an easy workaround for this to work, but I still would like it to work properly to avoid adding just another patch

Once I got around this error, I was able to take a .gif to show the `codemirrordialog` "in action", there are still style issues that I need to fix but I thought it would be a good idea to show it anyway.

As for the **next steps**, here is what we could do:

- For the moment only merge the plugin without the dialog code
- Fix issues and styles with dialog
- Ship it

But I'm going to wait for your comments and feedback

Thanks!


carloslancha:

julien:
Closing this in favor of #121 - will send another PR once #121 is merged with the dialog

PR #121, opened 2020-08-24
feat: create and use codemirror plugin

carloslancha:
Follow up of https://github.com/liferay/liferay-ckeditor/pull/120

- Removed preview dialog part
- Created infra to load codemirror and html-beautify as dependencies

~Problems:
When you try to change to source view you get the error `html_beautify is not defined`, which is being used in the plugin. That dependency was included in the previous minified distribution of codemirror so I guess we need to include it manually, but I'm not sure about the correct package.~

codemirror

Steps to test:

- Build `liferay-ckeditor`
- Copy / reference local `liferay-ckeditor` in `liferay-portal`
- Deploy `frontend-editor-ckeditor-web`
- Go to any editor in portal that uses ckeditor


julien:
Hey @carloslancha, 

Concerning the error you're getting, it's because I was also using the `js-beautify` package (which also provides `html_beautify` function) to format the "html" before passing it to the plugin.

This might not be needed, so I was going to look how `alloyeditor`'s source view handled that before sending the final PR.

plugins/codemirror/plugin.js
```diff
@@ -0,0 +1,130 @@
...
+				instance.codeMirrorEditor.on('change', instance._handleCodeMirrorChange.bind(instance, editor, oldData));
+
+				editor.on('resize', instance._handleEditorResize.bind(instance));
+
+				editor.fire('ariaWidget', this);
+
+				callback();
+			});
+
+			editor.addCommand('codemirror', codemirror.commands.source);
```
julien, line 40: @carloslancha this is my fault and I was going to change it in my previous PR:
I'm pretty sure this should go on line 100 (or at least in the `init` function). 

I'm not getting any error in dxp, but if I test our build of `ckeditor` in simple html page I'm getting this error because the `codemirror` command isn't defined in the `mode` event listener (declared on line 101)

error-01

Coud you please have a look? 
Thanks

carloslancha: Mmmm, shouldn't we move
```
if (editor.ui.addButton) {
				editor.ui.addButton('Source', {
					label: editor.lang.codemirror.source,
					command: 'codemirror',
					toolbar: 'mode,10'
				});
			}

			editor.on('mode', function() {
				editor.getCommand('codemirror').setState(
					editor.mode === 'source'
					? CKEDITOR.TRISTATE_ON
					: CKEDITOR.TRISTATE_OFF
				);
			});
``` 

to the end of `_createCodeMirrorEditor` so we're sure that everything is loaded and initialized?

julien: Yeah, that's actually a good idea.
Let me try that too though, just to be sure :+1:

julien: @carloslancha after looking at this, we should keep this in the `init` function, because if we call it in the `_createCodeMirrorEditor` function the "source" button won't be available in the toolbar as you can see in this screenshot

plugins/codemirror/plugin.js
```diff
@@ -0,0 +1,130 @@
...
+			if (editor.plugins.detectConflict('codemirror', ['sourcearea'])) {
+				return;
+			}
+
+			if (editor.elementMode === CKEDITOR.ELEMENT_MODE_INLINE) {
+				return;
+			}
+
+			if (!stylesLoaded) {
+				CKEDITOR.document.appendStyleSheet(
```
jbalsas, line 71: Could we use the `deps.json` list here? It will be easy enough to forget to load a dependency since we have to add it both there and here.

plugins/codemirror/deps.json
```diff
@@ -0,0 +1,10 @@
+{
+    "codemirror": [
+        "lib/codemirror.css",
+        "lib/codemirror.js",
+        "mode/xml/xml.js"
+    ],
+    "js-beautify": [
+        "js/lib/beautify-html.js"
+    ]
```
jbalsas, line 9: Based on https://github.com/liferay/liferay-ckeditor/pull/121/files#r476273541, if we wanted to go with that... we'd probably want to store `js` and `css` deps in different places so we could load them in different ways in an automatic way that doesn't rely on us in coding those paths twice.

plugins/codemirror/plugin.js
```diff
@@ -0,0 +1,130 @@
...
+
+			var instance = this;
+
+			CKEDITOR.scriptLoader.load([
+					instance.path + 'deps/codemirror/codemirror.js',
+					instance.path + 'deps/js-beautify/beautify-html.js'
+				],
+				function() {
+					CKEDITOR.scriptLoader.load(
+						instance.path + 'deps/codemirror/xml.js',
```
jbalsas, line 85: I'm assuming that adding `xml.js` right after `codemirror.js` will fail because it might run before it?

This is just a nit, but it feels oddly specific and makes this harder to abstract... one option would be to concat this with `codemirror.js` during build time, since that's what we always want and that way we'd also be avoiding additional requests.

jbalsas (approved):
Just added a couple of passing-by comments, but feel free to push this and do whatever you want with them :)

plugins/codemirror/plugin.js
julien: > I'm assuming that adding xml.js right after codemirror.js will fail because it might run before it?

Yes exactly, `xml.js` [assumes](https://github.com/codemirror/CodeMirror/blob/6fcc49d5321261b525e50f111f3b28a602f01f71/mode/xml/xml.js#L5-L10) there's a global `CodeMirror` object or a `require` function to be able to load it (not necessarily in that order)

During development, I played around with this and managed to use `ugligyjs` to concatenate and minify both files into one, so I also think it would be possible to concatenate (just with `cat` or similar since CKBuilder takes care of minifying) both files as part of the build and it should keep on working.

jbalsas: Could even cat `beautify` as well and have a single `vendor` file or something like that...

Anyways, this can be a small correction to apply once everything else is done.

carloslancha: That's actually a good one, we can create a single deps file concatenating all of them during the build process so we can load only one here.

plugins/codemirror/plugin.js
carloslancha: Not needed now that we're loading a single file :)

plugins/codemirror/deps.json
carloslancha: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FwzD3nQPA4gqHK%2Fgiphy.gif&token=done" data-gitmeme-token="done"></a>

carloslancha:
Updated the dependencies script so we create a single vendor file by type.

@julien could you please push a fix for what you commented [here](https://github.com/liferay/liferay-ckeditor/pull/121#discussion_r476222749) ?

plugins/codemirror/plugin.js
jbalsas: Nice!!

One could even go as far as to actually directly put the dependencies code on top of the plugin itself in the same file...

just-sayin

julien:
sorry @carloslancha I was busy with something else.
Concerning the fix, I could if I had access to your `liferay-ckeditor` repository, (I know that GitHub is annoying for that)

It's just a matter of putting the command "registration", i.e. this line 

```javascript
editor.addCommand('codemirror', CKEDITOR.plugins.codemirror.commands.source);
```

in the `init` function

julien:
@carloslancha I'm going to merge this  - I still need to add the dialog so I can fix that in the next PR 
Thanks

plugins/codemirror/plugin.js
carloslancha: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FfHgKZHDH1BHzi%2Fgiphy.gif&token=stitch" data-gitmeme-token="stitch"></a>

jbalsas: No need to do it, just saying!! 😉

jbalsas: whutchumean

carloslancha: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2F7XIq6GWNwOuUE%2Fgiphy.gif&token=forrest" data-gitmeme-token="forrest"></a>

PR #122, opened 2020-08-26
feat: add codemirror dialog

julien:
This is the second part of the [LPS-118290](https://issues.liferay.com/browse/LPS-118290) task:
The `codemirror` plugin has been updated and now includes a dialog that
can be used to edit and preview the code, when the editor is in `source`
mode

Here's a screenshot of the dialog in action:

codemirrordialog

Notes: 
The preview for the source code is rendered in an `iframe` to be able to load the `contentsCss` (if any) passed in the editor's `config` - without affecting the dialog itself.

### DIFF
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -40,8 +40,12 @@ CKEDITOR.dialog.add('codemirrordialog', function(editor) {
 			);
 
 			var preview = dialog.getContentElement('main', 'preview').getElement();
-			preview.setHtml(codeMirrorEditor.getValue());
 			preview.setSize('width', defaultWidth);
+
+			var iframe = this._createContentIframe(preview);
+			var iframeBody = iframe.$.contentDocument.body;
+
+			iframeBody.innerHTML = codeMirrorEditor.getValue();
 
 			codeMirrorEditor.on(
 				'change',
@@ -49,11 +53,89 @@ CKEDITOR.dialog.add('codemirrordialog', function(editor) {
 			);
 		},
 
+		_createContentIframe: function(parentElement) {
+			var dialog = this.dialog;
+			var editor = dialog.getParentEditor();
+
+			var mainElement = dialog.getContentElement('main').getElement();
+
+			var tabPanel = mainElement.getAscendant(function(el) {
+				return el.getName() === 'div' && el.getAttribute('role') === 'tabpanel';
+			});
+
+			var tabPanelParent = tabPanel.getParent();
+
+			var padding = {
+				bottom: parseInt(tabPanelParent.getComputedStyle('padding-bottom'), 10) || 0,
+				top: parseInt(tabPanelParent.getComputedStyle('padding-top'), 10) || 0
+			};
+
+			var height = tabPanel.getSize('height') - (padding.bottom + padding.top);
+
+			var iframe = new CKEDITOR.dom.element('iframe');
+			parentElement.append(iframe);
+
+			iframe.setAttributes({
+				'class': 'cke_wysiwyg_frame cke_reset',
+				'frameborder': 0,
+			});
+
+			iframe.setStyles({
+				'height': height + 'px',
+				'width': '99%'
+			});
+
+			var iframeDocument = iframe.$.contentDocument;
+			var iframeHead = iframeDocument.head;
+
+			var contentsCss = editor.config.contentsCss;
+
+			if (Array.isArray(contentsCss)) {
+				contentsCss.forEach(function(url) {
+					var link = iframeDocument.createElement('link');
+					link.setAttribute('href', url);
+					link.setAttribute('rel', 'stylesheet');
+
+					iframeHead.appendChild(link);
+				});
+			}
+			else {
+				var link = iframeDocument.createElement('link');
+				link.setAttribute('href', contentsCss);
+				link.setAttribute('rel', 'stylesheet');
+
+				iframeHead.appendChild(link);
+			}
+
+			var direction = editor.config.contentsLangDirection;
+
+			var iframeHtml = iframeDocument.documentElement;
+			iframeHtml.setAttribute('dir', direction);
+			iframeHtml.setAttribute('lang', editor.config.defaultLanguage);
+
+			var iframeBody = iframeDocument.body;
+			iframeBody.classList.add('cke_editable');
+			iframeBody.classList.add('cke_editable_themed');
+			iframeBody.classList.add('cke_contents_' + direction);
+
+			iframeBody.setAttribute('contenteditable', false);
+			iframeBody.setAttribute('spellcheck', false);
+
+			iframeBody.style.background = '#fff';
+
+			return iframe;
+		},
+
 		_handleCodeMirrorChange: function() {
 			var newData = this.codeMirrorEditor.getValue();
 			var preview = this.dialog.getContentElement('main', 'preview').getElement();
 
-			preview.setHtml(newData);
+			var iframe = preview.findOne('iframe');
+			if (iframe && iframe.$) {
+				var iframeDocument = iframe.$.contentDocument;
+				var iframeBody = iframeDocument.body;
+				iframeBody.innerHTML = newData;
+			}
 		},
 
 		contents: [{
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -1,0 +1,108 @@
+CKEDITOR.dialog.add('codemirrordialog', function(editor) {
+	var editorWindow = CKEDITOR.document.getWindow();
+	var size = editorWindow.getViewPaneSize();
+	var scalex = 0.9;
+	var scaley = 0.7;
+	var height = size.height * scaley;
+	var width = size.width * scalex;
+
+	if (!editor.window) {
+		editor.window = editorWindow;
+	}
+
+	return {
+		_createCodeMirrorEditor: function() {
+			var dialog = this.dialog;
+
+			var size = dialog.getSize();
+
+			var textarea = dialog.getContentElement('main', 'data')
+				.getInputElement().$;
+
+			var codeMirrorEditor = this.codeMirrorEditor =
+				CodeMirror.fromTextArea(
+					textarea,
+					{
+						lineNumbers: true,
+						lineWrapping: true,
+						mode: 'text/html'
+					}
+				);
+
+			var defaultWidth = (size.width * 0.5) - 10;
+
+			codeMirrorEditor.setSize(defaultWidth, null);
+
+			var editor = dialog.getParentEditor();
+
+			codeMirrorEditor.setValue(
+				html_beautify(editor.getData(true))
+			);
+
+			var preview = dialog.getContentElement('main', 'preview').getElement();
+			preview.setHtml(codeMirrorEditor.getValue());
+			preview.setSize('width', defaultWidth);
+
+			codeMirrorEditor.on(
+				'change',
+				this._handleCodeMirrorChange.bind(this)
+			);
+		},
+
+		_handleCodeMirrorChange: function() {
+			var newData = this.codeMirrorEditor.getValue();
+			var preview = this.dialog.getContentElement('main', 'preview').getElement();
+
+			preview.setHtml(newData);
+		},
+
+		contents: [{
+			id: 'main',
+			elements: [{
+				align: 'top',
+				children: [
+					{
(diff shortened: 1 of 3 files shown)


jbalsas:
Looking good feature-wise!! 👍

carloslancha:
As we talked offline there's a problem with the preview. I suppose that what we want to show in the preview is exactly the same we would be watching in the regular content editor, so all the `contentCss` styles should be loaded there to style the preview in the same way.

But we can't do that unless the preview content is inside an iframe (in the same way the regular content editor works), we should not load `contentCss` style files in the page body because it would affect the entire page. So

- If possible we should place the ckeditor regular content editor as the preview (but in readonly mode) and update the content from the editor with `editor.setData' 
- If not possible we should mimic ckeditor regular content editor in the preview, wrapping it inside an iframe, loading there the `contentCss` files and placing some custom js that would listen to editor data changes and update the content.

jbalsas:
> But we can't do that unless the preview content is inside an iframe (in the same way the regular content editor works), we should not load contentCss style files in the page body because it would affect the entire page. So

But if the window is in the same page as the styles... then it should be actually picking up the styles, shouldn't it? Unless it's picking other styles.

carloslancha:
> But if the modal is in the same page as the styles... then it should be actually picking up the styles, shouldn't it? Unless it's picking other styles.

You can configure the editor via `ConfigContributor` passing the files you want to load with `contentCss` which are only loaded in the content editor, so if you pass a file that is not in the page (because you only want it to be loaded inside the content editor) then no, it won't be picking them.

jbalsas:
> [...] so if you pass a file that is not in the page (because you only want it to be loaded inside the content editor) then no, it won't be picking them.

Not sure I understand this... where is the modal being opened?
- In the top window. Shouldn't the original styles be applying there?
- Inside the ckeditor iframe. Shouldn't the `contentCSS` styles be applying there?
- In a different iframe. For sure we won't have styles here

julien:
@jbalsas the modal is being opened in the top window, but we're not getting the styles loaded in the "editor area" (i.e. the place where we actually enter text) which is in a `iframe`. 

@carloslancha one difficult thing to do which I'm looking at right now, is cloning the `iframe` to use it in the dialog instead of my "preview" `div`. One problem is that when you switch from "wysiwyg" to "source" mode, that `iframe` is already gone, so before opening the dialog I would need to clone the `iframe` and somehow pass it to the dialog from the plugin ... yes I know this sounds crazy.

carloslancha:
> Not sure I understand this... where is the modal being opened?

In the top window.

> I mean, you're saying we're passing additional contentCss css that is not applied in the admin page?

I'm saying there's an API that gives you the possibility to do it.

Also, if we need to add 
```
.cke_dialog .code_preview hr {
	border: none;
	background-color: #ccc;
	color: #ccc;
	height: 1px;
}
```

to the plugin styles because if not `hr` will not be visible, we're obviously not mimicking the editor view

jbalsas:
So... let me see if I understand it:

Screen_Shot_2020-08-26_at_15_38_28

My question is... why is there an iframe there at all? Do all CKEditor modals use an iframe? (that's unexpected, but who am I to judge...)

jbalsas:
> Also, if we need to add
>
> .cke_dialog .code_preview hr {
>	border: none;
>	background-color: #ccc;
>	color: #ccc;
>	height: 1px;
>}

Yes... please, let's not ever ever do this again and play with styles and content. Let's burn this in our heads. Maybe even add a disclaimer or a rule or something in this repo.

Now, where I wanted to go is... if there's no `iframe`, there should be no problem. At least in the same sense there's no problem right now with the current implementation of this plugin for AlloyEditor. So maybe we could try to get rid of the `iframe` rather than trying to put things inside the `iframe`.

carloslancha:
Right now there is **no** iframe in the plugin. CKEditor content editor **is** an iframe where `contentCss` is loaded.

The `contentCss` loaded in the content editor iframe doesn't necessary have to be the same styles loaded in the page. If what we want is the preview looks exactly equal to the content editor I think we should load and isolate it in the same way, if not it may work, or not.

If we're happy with this affirmation... well, @julien the problem with the `<hr>` it's probably because you're rendering the dialog with (or inside a container with) the class `cke_reset_all`, so you need to render it outside it (or without the class `cke_reset_all`)

jbalsas:
> The contentCss loaded in the content editor iframe doesn't necessary have to be the same styles loaded in the page. If what we want is the preview looks exactly equal to the content editor I think we should load and isolate it in the same way, if not it may work, or not.

Yes, I think that's a valid compromise we can make. I'd expect this to work initially at least like AlloyEditor's implementation... which I'm pretty sure didn't cover this either. The thing with AlloyEditor was that it basically assumed the environment it was surrounded by. That's the main usage of `contentCSS`, to basically mimic the things outside the content iframe. 

You're 100% correct that it is not technically the same and there could be cases where the `contentCSS` would not match the outer window css. 

For our case, that's exactly what we attempt with `contentCSS`... to match what's outside (and style CKEditor UI, maybe?, which does not play in the equation). So, I feel that loosening up this requirement here should get us to where we want to get faster. 

**If** at some point we indeed see the need for the `preview` to receive different styles than the parent window, then surely we'll have to go with the iframe+styles approach... which will probably look worse UX-wise and will have more edge cases to cover... 

@julien, please work on the styles, avoid adding CSS that modifies how the content looks like and figure out what are the differences between the same content in our preview modal and outside to get them match as much as possible. Compare with the current AlloyEditor behaviour and when we're close enough, we should be good to go.

julien:
> If we're happy with this affirmation... well, @julien the problem with the <hr> it's probably because you're rendering the dialog with (or inside a container with) the class cke_reset_all, so you need to render it outside it (or without the class cke_reset_all)

@carloslancha I'm only creating a dialog command and CKEditor does the rest , so if I want to get rid of the  `cke_reset_all` class applied to the dialog, the only way I have of doing it, is via DOM manipulation somewhere in my dialog code, and that doesn't sound right to me.

> please work on the styles, avoid adding CSS that modifies how the content looks like
@jbalsas I don't understand: if I have to work on the styles I'll probably have to add CSS or modify it.

jbalsas:
> @jbalsas I don't understand: if I have to work on the styles I'll probably have to add CSS or modify it.

Working on the styles means figuring out why the preview doesn't look like it's on a page. If to do that you have to remove the `cke_reset_all`... then maybe that's what you have to do and figure out the best way to do it :). 

**You must not add any custom CSS to make the preview look okay because that CSS does not apply when the dialog is not present or the content is in a page without the editor**

carloslancha:

julien:
This is pretty tricky to be honest: removing `cke_reset_all` from the dialog makes the dialog look worse, but I've got something that's more or less reasonable

Here's a .gif of what it looks like in our `ckeditor` build 

ckeditor-standalone

And here's how it looks in `liferay-portal`

ckeditor-portal

julien:
@jbalsas and @carloslancha could you tell me when you have some time what you think about the changes?
Thanks

jbalsas:
Hey @julien, to be honest I'm a bit confused about the last one... I assume we now render the content inside an `iframe` so that CKEditor styles don't clash... and add all the `contentCSS`... sounds a bit convoluted, but if it works... 👍

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -40,20 +40,102 @@ CKEDITOR.dialog.add('codemirrordialog', function(editor) {
...
+			var direction = editor.config.contentsLangDirection;
+
+			var iframeHtml = iframeDocument.documentElement;
+			iframeHtml.setAttribute('dir', direction);
+			iframeHtml.setAttribute('lang', editor.config.defaultLanguage);
+
+			var iframeBody = iframeDocument.body;
+			iframeBody.classList.add('cke_editable');
+			iframeBody.classList.add('cke_editable_themed');
+			iframeBody.classList.add('cke_contents_' + direction);
```
jbalsas, line 119: Why do you need these CKEditor classes? If you're doing a preview, there should be no trace at all of CKEditor.

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -40,20 +40,102 @@ CKEDITOR.dialog.add('codemirrordialog', function(editor) {
...
+
+			iframeBody.setAttribute('contenteditable', false);
+			iframeBody.setAttribute('spellcheck', false);
+
+			iframeBody.style.background = '#fff';
```
jbalsas, line 124: Is this needed? Isn't this coming from the `contentCss`?

jbalsas (approved):
Just some minor comments... if this works and you think it's better than the other approach, I'd say ship it. We can continue tweaking it next week.

@carloslancha was right all along ;)

carloslancha:
Hey @julien could you please update the description or add a comment with what we talked yesterday about why this is the solution? Thx!!

julien:
Thanks for the feedback!

@jbalsas you're right we're rendering in an `iframe`:

I tried removing the `cke_reset` class from the dialog but as I [said](https://github.com/liferay/liferay-ckeditor/pull/122#issuecomment-681860367) that breaks everything, by using an iframe an loding the `contentsCss` we get a better result.

@carloslancha yes, I'm going to update the description once I've fixed what's needed :+1:

plugins/codemirror/dialogs/codemirrordialog.js
julien: You're right, I'm going to remove those and see what happens

plugins/codemirror/dialogs/codemirrordialog.js
julien: Yes, that's a good point.

In a "vanilla" environment, i.e. when using our build of `liferay-ckeditor` in html page that doesn't have any customization (no custom "editor" configuration) the `background-color` of the `body` inside the `iframe` is not set; but when using it in `liferay-portal` the background color was set to light grey and I'm thinking that it might be related to what you mentioned in [this](https://github.com/liferay/liferay-ckeditor/pull/122#discussion_r478861646) previous comment, so I'm also going to try to remove this and see if that fixes it.

carloslancha: AFAIK `cke_contents` class adds a grey background, that's part of the `moono-lexicon` skin, so if you're adding `contentCss` files you're adding (in portal) `moono-lexicon` too, so this is needed to get rid of that style. 

Other option could be add a class `cke_preview` and overwrite that background from a css in the skin, but the result would be the same

julien: @carloslancha yes that's right.

So if I remove all those classes from the `iframe` and it's `body` this is what it looks like 

no-classes-in-iframe-or-body

So I don't really think the "padding" is a big issue but the background does add more noise.

> Other option could be add a class cke_preview and overwrite that background from a css in the skin, but the result would be the same

I tried that too, but I was unable to make it work because another class is adding the background color to the body, and even with `background: #fff !important` it wasn't getting applied. I agree that the setting the background via "style" is a bit ugly but it works

jbalsas: ok, let's keep it as it is for now and ship it... it's already better than what we have :)

carloslancha: yep, you probably need to be more specific in the class `.cke_dialog.cke_preview .cke_contents .cke_editable` and `!important` or something like that, but as Chema said, is not a big issue 😄

carloslancha:
LGTM!

We can make some improvements here (but not know), maybe we can create issues to see if it's worth the time at some moment
- Replace inline styling with css in the plugin styles
- Find out how to improve the icons css classes generation to support plugins so we don't need to convert the icons and place them manually

Nice job @julien !

carloslancha:
@julien could you please take care of merging this when ready and prepare a release?

<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia3.giphy.com%2Fmedia%2FE3cQDsj7pUpwI%2Fgiphy.gif&token=kraken" data-gitmeme-token="kraken"></a>

julien:
@jbalsas, @carloslancha thanks!

Yes I agree that we can improve this along the way and add some tasks for the icon generation (i.e. separate "core" icons from "plugin specific" icons, which is something we talked about with @carloslancha)

I updated the description just to mention we were rendering the preview in an `iframe` to be able to load `contentsCss`, so I'll merge it and prepare a release 

Speaking of which, there's probably some "work" to be done when integrating in `liferay-portal`: i.e. when we're using `bbcode` or `creole` we probably want to use the default `sourcearea` plugin but I guess that's going to be discussed in another PR.

PR #124, opened 2020-08-26
fix: prevent iconsClassesGenerator from crashing

### LINKED ISSUE #123: Prevent `iconsClassesGenerator.js` from failing if icons aren't found
The [`iconsClassesGenerator.js`](https://github.com/liferay/liferay-ckeditor/blob/20e6ce2c1849e213be42664bf2a5523f73d87b86/support/iconsClassesGenerator.js) script [reads](https://github.com/liferay/liferay-ckeditor/blob/20e6ce2c1849e213be42664bf2a5523f73d87b86/support/iconsClassesGenerator.js#L106-L126) files but if these files are non existent it will crash and prevent the build script from generating the `ckeditor` bundle.

Here's how to make it crash:

- Edit [icons.json](https://github.com/liferay/liferay-ckeditor/blob/20e6ce2c1849e213be42664bf2a5523f73d87b86/skins/moono-lexicon/icons/icons.json) and add a new "entry" to the `icons` object, for example 

    ```json
    ...
    icons: {
       "preview": "preview"
    }
    ...
    ```

- Build `ckeditor` with 

    ```sh
    ./ck.sh build
    ```

- See it fail with the following message:

    ```js
    fs.js:115
        throw err;
        ^

    Error: ENOENT: no such file or directory, open '../node_modules/@clayui/css/lib/images/icons/preview.svg'
        at Object.openSync (fs.js:439:3)
        at Object.readFileSync (fs.js:344:35)
        at Object.<anonymous> (/home/jc/Documents/liferay-ckeditor/support/iconsClassesGenerator.js:106:22)
        at Module._compile (internal/modules/cjs/loader.js:689:30)
        at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
        at Module.load (internal/modules/cjs/loader.js:599:32)
        at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
        at Function.Module._load (internal/modules/cjs/loader.js:530:3)
        at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
        at startup (internal/bootstrap/node.js:283:19)
    ```

julien:
Closes #123

### DIFF
--- a/support/iconsClassesGenerator.js
+++ b/support/iconsClassesGenerator.js
@@ -102,30 +102,53 @@ for (const [cKEditorIcon, clayIcon] of Object.entries(iconsConfig.icons)) {
 	let directionClass = '';
 	let svgDataList = [];
 
-	if (typeof clayIcon === 'string') {
-		const svgData = fs.readFileSync(
-			`${sourceIconsPath}/${clayIcon}.svg`,
-			'utf8'
-		);
+	let svgData;
 
-		iconsCSSContent += getCSS(svgData, cKEditorIcon);
-	} else {
-		if (clayIcon.ltr) {
-			const svgData = fs.readFileSync(
-				`${sourceIconsPath}/${clayIcon.ltr}.svg`,
+	if (typeof clayIcon === 'string') {
+		try {
+			svgData = fs.readFileSync(
+				`${sourceIconsPath}/${clayIcon}.svg`,
 				'utf8'
 			);
 
-			iconsCSSContent += getCSS(svgData, cKEditorIcon, 'ltr');
+			iconsCSSContent += getCSS(svgData, cKEditorIcon);
+		} catch (err) {
+			console.warn(
+				'Could not read:',
+				`${sourceIconsPath}/${clayIcon}.svg`
+			);
+		}
+	} else {
+		if (clayIcon.ltr) {
+			try {
+				svgData = fs.readFileSync(
+					`${sourceIconsPath}/${clayIcon.ltr}.svg`,
+					'utf8'
+				);
+
+				iconsCSSContent += getCSS(svgData, cKEditorIcon, 'ltr');
+			} catch (err) {
+				console.warn(
+					'Could not read:',
+					`${sourceIconsPath}/${clayIcon.ltr}.svg`
+				);
+			}
 		}
 
 		if (clayIcon.rtl) {
-			const svgData = fs.readFileSync(
-				`${sourceIconsPath}/${clayIcon.rtl}.svg`,
-				'utf8'
-			);
+			try {
+				svgData = fs.readFileSync(
+					`${sourceIconsPath}/${clayIcon.rtl}.svg`,
+					'utf8'
+				);
 
-			iconsCSSContent += getCSS(svgData, cKEditorIcon, 'rtl');
+				iconsCSSContent += getCSS(svgData, cKEditorIcon, 'rtl');
+			} catch (err) {
+				console.warn(
+					'Could not read:',
+					`${sourceIconsPath}/${clayIcon.rtl}.svg`
+				);
+			}
 		}
 	}
 }
--- a/support/iconsClassesGenerator.js
+++ b/support/iconsClassesGenerator.js
@@ -99,9 +99,6 @@ const getCSS = (svgData, cKEditorIcon, direction) => {
 let iconsCSSContent = '';
 
 for (const [cKEditorIcon, clayIcon] of Object.entries(iconsConfig.icons)) {
-	let directionClass = '';
-	let svgDataList = [];
-
 	let svgData;
 
 	if (typeof clayIcon === 'string') {


support/iconsClassesGenerator.js
```diff
@@ -102,30 +102,53 @@ for (const [cKEditorIcon, clayIcon] of Object.entries(iconsConfig.icons)) {
 	let directionClass = '';
 	let svgDataList = [];
 
-	if (typeof clayIcon === 'string') {
-		const svgData = fs.readFileSync(
-			`${sourceIconsPath}/${clayIcon}.svg`,
-			'utf8'
-		);
+	let svgData;
```
carloslancha, line 105: Why extracting this?

carloslancha, line 105: Why extracting this?

carloslancha (approved):
LGTM. Just left one non blocking comment

support/iconsClassesGenerator.js
julien: Why have 3 `const` when you can have one `let`

jbalsas: Since you're going this way... why have `let` at all? just inline the calls, you're not using `svgData` for anything else...

Looking on a quick look, I think you could even do something like:

```javascript
const direction = typeof clayIcon === 'object' || clayIcon.rtl ? 'rtl' : 'ltr';
const iconPath = clayIcon === 'string' ? clayIcon : clayIcon.ltr | clayIcon.rtl;

iconCssContent += getCss(
    fs.readFileSync(`${sourceIconsPath/${iconPath}}.svg`, 'utf8'),
    ckEditorIcon,
    direction
);
```

Not saying this reads better, but it's probably as hard to read as the one we have right now because all the code looks so similar that it's hard to tell the branches apart.

julien: The thing is that when `clayIcon` is an object, it usually has [both](https://github.com/liferay/liferay-ckeditor/blob/09632a29a40f92c9bf5088d166520873f81125e3/skins/moono-lexicon/icons/icons.json#L15-L30) `ltr` and `rtl` properties, and we want to generate the class for both so this in this loop

```js
const direction = typeof clayIcon === 'object' || clayIcon.rtl ? 'rtl' : 'ltr';
const iconPath = clayIcon === 'string' ? clayIcon : clayIcon.ltr | clayIcon.rtl;
```

won't do it.

One thing I'll can do is get rid of the [unused](https://github.com/liferay/liferay-ckeditor/blob/09632a29a40f92c9bf5088d166520873f81125e3/support/iconsClassesGenerator.js#L102-L103) variables in the loop

jbalsas: You're totally right 👍

jbalsas: But let's not touch this too much :)

julien: @jbalsas yes that would even be better. 
I agree, I'm going to focus on #122 right now, and when the time comes we can improve or clean up these helpers scripts.

wincent:
I don't really get this. [#123](https://github.com/liferay/liferay-ckeditor/issues/123) says that if you configure `icons.json` to refer to a non-existent file, the build will crash; but isn't that a feature, not a bug?

There's a risk that any crash that we `catch` in here and which prints a warning in the console will just go unnoticed among all the other build output. Is there some other back-story here that I am missing?

julien:
@wincent I don't know if it's a bug or a feature, but I think there's something we could improve.

Let's assume the [`icons.json`](https://github.com/liferay/liferay-ckeditor/blob/20e6ce2c1849e213be42664bf2a5523f73d87b86/skins/moono-lexicon/icons/icons.json) file has an "invalid' entry"; you then run the build script and the skin's directory is [copied](https://github.com/liferay/liferay-ckeditor/blob/20e6ce2c1849e213be42664bf2a5523f73d87b86/ck.sh#L62) to the `ckeditor-dev/skin` directory, when the build tries to generate the "icon classes" it will fail abort the build - so you edit `icons.json` and remove the "invalid entry" you had added before, you re-run the build and it fails again ...

So the "fix" that's probably more reasonable would be to change [this](https://github.com/liferay/liferay-ckeditor/blob/20e6ce2c1849e213be42664bf2a5523f73d87b86/ck.sh#L62) line from 

```sh
```

to

```sh
cp -r skins/moono-lexicon ckeditor-dev/skins/
``` 

Does it make more sense now?

carloslancha:
Just as an addition to what @julien said, the problem is that if the build fails we don't reach the script line where we're removing the copied skin from `ckeditor-dev`.

What @wincent says makes sense, it's preferable that the build crashes on this kind of errors, it's to easy to miss the error message.

I'm ok with the solution @julien proposes about changing how the skin is copied, that way we'll be sure that everything is being overwrite. It could help that we remove the skin **before** copying it if the folder exists just in case?

wincent:
> Does it make more sense now?

I don't quite get this bit:

> you edit `icons.json` and remove the "invalid entry" you had added before, you re-run the build and it fails again ...

Why will the build fail on the second run? I get that it might be invalid though in the sense that if you run `cp -r a b/a` twice in a row then the first time it will make a copy of `a` at `b/a`, and the second time it will put _another_ copy of `a` inside `b/a`, so changing it to `cp -r skins/moono-lexicon ckeditor-dev/skins/` makes sense.

To @carloslancha's point:

> It could help that we remove the skin before copying it if the folder exists just in case?

Yes, or we could use `rsync`:

```
rsync -a --delete skins/moono-lexicon ckeditor-dev/skins
```

which will copy missing/changed files and remove no-longer-used files.

julien:
@wincent, the second time I run the build, even if I change the `icons.json` in `skins/moono-lexicon/icons/icons.json` file the  `ckeditor-dev/skins/moono-lexicon/icons/icons.json` file doesn't get updated

wincent:
> @wincent, the second time I run the build, even if I change the `icons.json` file in `skins/moono-lexicon/icons/icons.json` the `ckeditor-dev/skins/moono-lexicon/icons/icons.json` file doesn't get updated

I expect because:

- instead of writing to `ckeditor-dev/skins/moono-lexicon/icons/icons.json`

- it writes to `ckeditor-dev/skins/moono-lexicon/moono-lexicon/icons/icons.json`

Right?

julien:
@wincent Correct

carloslancha:
This should be fixed with https://github.com/liferay/liferay-ckeditor/pull/128 since we're creating a fresh copy of `ckeditor-dev` on each build.

PR #126, opened 2020-08-28
feat: add imagespacingbox plugin

mateomustapic:
This PR is a follow up on https://github.com/liferay-frontend/liferay-portal/pull/313

The purpose of these changes is adding the possibility for user to add horizontal and vertical space to the image element in editor.
I achieved this by creating `imagespacingbox` plugin which adds 2 additional input fields in `image2` plugin dialog box.
Setting `hspace` value adds horizontal margin, `margin-left` and `margin-right` to image element selected in editor.
Setting `vspace` value adds vertical margin, `margin-left` and `margin-right` to image element selected in editor.

This solves the issue when editing the text next to image, where there was previously no space and allows the user possibility to add custom space replicating functionality we had with `image` plugin.

Additional modifications made in this PR compared to https://github.com/liferay-frontend/liferay-portal/pull/313
- Used `setStyles()` to avoid calling `setStyle()` twice
- Used translation keys from `image` plugin

To test these changes, you can add image to any CKeditor instance in the portal:

For example
Go to
- Build ckeditor with with plugin changes
- Content and Data
- Web Content
- New Basic Web Content
- Upload Image to main editor
- Double click image to open dialog box with `Image properties`
- Add preferred `hspace` or `vspace` values

1. Add 12 hspace value

2. Image with text on the right without horizontal space

3. Image with text on the right with applied `hspace` value

### DIFF
--- a/ck.sh
+++ b/ck.sh
@@ -78,6 +78,9 @@ case "$COMMAND" in
 					node ../support/copyPluginDependencies.js $pluginDir
 				done
 
+				# Copy lang files from image plugin to imagespacingbox
+				cp -r plugins/image/lang plugins/imagespacingbox
+
 				# Generate SVG icons CSS Classes in custom skins
 				skinsWithIcons=$(find skins -maxdepth 2 -mindepth 2 -type f -name icons.json)
 				for iconsFile in $skinsWithIcons ; do
@@ -92,7 +95,7 @@ case "$COMMAND" in
 						--leave-css-unminified --leave-js-unminified --no-ie-checks
 				else
 					dev/builder/build.sh --build-config ../../../support/build-config.js \
-						--no-ie-checks 
+						--no-ie-checks
 				fi
 
 				# Remove old build files.
--- a/plugins/imagespacingbox/plugin.js
+++ b/plugins/imagespacingbox/plugin.js
@@ -1,0 +1,104 @@
+(function () {
+	'use strict';
+	CKEDITOR.plugins.add('imagespacingbox', {
+		init: function (editor) {
+			CKEDITOR.on('dialogDefinition', function (event) {
+				var dialogDefinition = event.data.definition;
+
+				var dialog = event.data.dialog;
+
+				if (dialog.getName() === 'image2') {
+					var infoTab = dialogDefinition.getContents('info');
+
+					if (infoTab.get('spacingBox')) {
+						return;
+					}
+
+					var spacingBox = {
+						children: [
+							{
+								children: [
+									{
+										commit: function (widget) {
+											widget.setData(
+												'hspace',
+												this.getValue()
+											);
+
+											var hspace = widget.data.hspace;
+
+											var imageElement =
+												widget.parts.image;
+
+											if (imageElement && hspace) {
+												imageElement.setStyles({
+													'margin-left':
+														hspace + 'px',
+													'margin-right':
+														hspace + 'px',
+												});
+											}
+										},
+										id: 'hspace',
+										label: editor.lang.image.hSpace,
+										requiredContent:
+											'img{margin-left,margin-right}',
+										setup: function (widget) {
+											this.setValue(widget.data.hspace);
+										},
+										type: 'text',
+										validate: CKEDITOR.dialog.validate.integer(
+											editor.lang.image.validateHSpace
+										),
+									},
+								],
+								type: 'hbox',
+							},
+							{
+								children: [
+									{
+										commit: function (widget) {
+											widget.setData(
+												'vspace',
+												this.getValue()
+											);
+
+											var vspace = widget.data.vspace;
+
+											var imageElement =
+												widget.parts.image;
+
+											if (imageElement && vspace) {
+												imageElement.setStyles({
+													'margin-bottom':
+														vspace + 'px',
+													'margin-top': vspace + 'px',
+												});
+											}
+										},
+										id: 'vspace',
+										label: editor.lang.image.vSpace,
+										requiredContent:
+											'img{margin-top,margin-bottom}',
+										setup: function (widget) {
+											this.setValue(widget.data.vspace);
+										},
+										type: 'text',
+										validate: CKEDITOR.dialog.validate.integer(
+											editor.lang.image.validateVSpace
+										),
+									},
+								],
+								type: 'hbox',
+							},
+						],
+						id: 'spacingBox',
+						type: 'hbox',
+					};
+
+					infoTab.add(spacingBox);
+				}
+			});
+		},
+	});
+})();
--- a/support/build-config.js
+++ b/support/build-config.js
@@ -45,6 +45,7 @@ var CKBUILDER_CONFIG = {
 		format: 1,
 		horizontalrule: 1,
 		image2: 1,
+		imagespacingbox: 1,
 		indentlist: 1,
 		indentblock: 1,
 		justify: 1,


carloslancha:

carloslancha:
Looks like we're missing some code formatting here, we need to add `plugins` folder to the format scripts

mateomustapic:
> Looks like we're missing some code formatting here, we need to add `plugins` folder to the format scripts
> 
> Edit: Looks like if we just add plugins to the prettier script it will not work due to some fixes CKEditor doesn't support. I'll open an issue to take care of it.

ok, thanks

carloslancha:
@mateomustapic  In the meantime, please, add this to the `package.json`

```
"format": "prettier --write \"(plugins|skins)/**/*.css\" \"(plugins|support)/**/*.js\" \"*.json\" \"*.md\"",
```

in the `.pretierrc.json`

```
"trailingComma": "none",
```

and run `yarn format`. That will format your plugin (and some other stuff, just ignore them). Try to build CKEditor, I think it should work and we'll have this plugin code with at least some minimal format

mateomustapic:
> @mateomustapic In the meantime, please, add this to the `package.json`
> 
> ```
> "format": "prettier --write \"(plugins|skins)/**/*.css\" \"(plugins|support)/**/*.js\" \"*.json\" \"*.md\"",
> ```
> 
> in the `.pretierrc.json`
> 
> ```
> "trailingComma": "none",
> ```
> 
> and run `yarn format`. That will format your plugin (and some other stuff, just ignore them). Try to build CKEditor, I think it should work and we'll have this plugin code with at least some minimal format

Done, I will now push the formatted plugin code

carloslancha:
@mateomustapic maybe I didn't explain well. Just change the package.json and the .prettierrc locally, run `yarn format` and push the changes on your plugin. Do not push any other changes, nor `package.json` nor `.pretierrc`

mateomustapic:
> @mateomustapic maybe I didn't explain well. Just change the package.json and the .prettierrc locally, run `yarn format` and push the changes on your plugin. Do not push any other changes, nor `package.json` nor `.pretierrc`

yeah I got it, but thought build was gonna break if I don't commit those files.
Anyway, I pushed now with only plugin code.

plugins/imagespacingbox/plugin.js
```diff
@@ -0,0 +1,103 @@
...
+										commit: function (widget) {
+											widget.setData(
+												'hspace',
+												this.getValue()
+											);
+
+											var hspace = widget.data.hspace;
+
+											var imageElement =
+												widget.parts.image;
```
carloslancha, line 32: Maybe add here a separation line?

plugins/imagespacingbox/plugin.js
```diff
@@ -0,0 +1,103 @@
...
+										commit: function (widget) {
+											widget.setData(
+												'vspace',
+												this.getValue()
+											);
+
+											var vspace = widget.data.vspace;
+
+											var imageElement =
+												widget.parts.image;
```
carloslancha, line 69: Maybe add here a separation line?

carloslancha (commented):
Code LGTM, just added two formatting comments. Thx!

plugins/imagespacingbox/plugin.js
mateomustapic: sure, good idea

plugins/imagespacingbox/plugin.js
```diff
@@ -0,0 +1,105 @@
+(function () {
+	'use strict';
+	CKEDITOR.plugins.add('imagespacingbox', {
+		requires: 'image',
```
julien, line 4: I think there's one thing we might be able to improve.

Here we `require` the `image` plugin because bellow [we're using](https://github.com/liferay/liferay-ckeditor/pull/126/files#diff-66f649f4e3ad8a5239ef948956905454R44) it's "language keys";  which means an additional request will be made, because since we [switched](https://github.com/liferay/liferay-ckeditor/pull/119) to the `image2` plugin, the `image` plugin is not bundled with `ckeditor.js` anymore.

You'd think that this additional request is made to `image.js` but since that plugin is not compatible with `image2.js`, because they have shared functionality (and mostly because of [this](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/plugins/image/plugin.js#L22-L26) line), what happens is that the request is made to `image2.js` which somehow doesn't make a lot of sense since `image2.js` is bundled in `ckeditor.js` 

It's more or less visible in the devtools, 

1. We `require` the `image` plugin but since it's not compatible with our bundled `image2` plugin we get this warning

error01

2. So `image2` ends up being "required" 

error02

3. Which isn't really needed since `image2` is already bundled in our `ckeditor.js` build

error03

Ideally we probably want to copy existing `lang` directories/files that contain translated language keys to our "custom" plugins' directories or reuse language keys that are available

mateomustapic: hey @julien 
This was the only way I manage to use `image` plugin's translation keys, but I see it is not the right way to do it.
I will now work on copying `lang` files to `imagespacingbox` and push the code as soon as I can.

mateomustapic: Ok, so I copied language keys from `image` plugin to `plugins/imagespacingbox` to avoid using `requires:`image`` as before. 
I have built ckeditor, these keys are after build in `ckeditor-dev/plugins/imagespacingbox` directory and working as expected in the portal. 
lang

julien: @mateomustapic ok that seems better :+1: 

However, everyone needs to be able to do this. So you probably need to edit `ck.sh` to make sure the `lang` directory from `ckeditor-dev/plugins/image` is copied to `plugins/imagespacingbox` before running the build 

Does that make sense?

mateomustapic: Yes, makes sense. 
Just wanted to confirm if this is what we want to achieve.

julien:
@mateomustapic this looks better.
Just one small thing, could you check the formatting?

When looking at `plugin.js` some things look a bit wrong 

image

On line [31](https://github.com/liferay/liferay-ckeditor/pull/126/files#diff-66f649f4e3ad8a5239ef948956905454R31) couldn't `widget.part.image` be on the same line? Same thing for line [69](https://github.com/liferay/liferay-ckeditor/pull/126/files#diff-66f649f4e3ad8a5239ef948956905454R69)

mateomustapic:
> @mateomustapic this looks better.
> Just one small thing, could you check the formatting?
> 
> When looking at `plugin.js` some things look a bit wrong
> 
> image
> 
> On line [31](https://github.com/liferay/liferay-ckeditor/pull/126/files#diff-66f649f4e3ad8a5239ef948956905454R31) couldn't `widget.part.image` be on the same line? Same thing for line [69](https://github.com/liferay/liferay-ckeditor/pull/126/files#diff-66f649f4e3ad8a5239ef948956905454R69)

I added `"format": "prettier --write \"(plugins|skins)/**/*.css\" \"(plugins|support)/**/*.js\" \"*.json\" \"*.md\""` to `package.json` as Carlos previously requested, to format plugins directory and this is the formatted version.

julien:
@mateomustapic ok then forget about that comment.
I'm going to test in portal quickly just to double check

julien:
@mateomustapic so this is working, but I thought we were going to add more input fields to be able to add top, right, bottom and left margins independently?

If nothing about that has been decided and we want to be able to do what the `image` plugin used to do then this is fine.

mateomustapic:
> @mateomustapic so this is working, but I thought we were going to add more input fields to be able to add top, right, bottom and left margins independently?
> 
> If nothing about that has been decided and we want to be able to do what the `image` plugin used to do then this is fine.

In the last discussion we had, the idea was to do it this way in `liferay-ckeditor` and merge it as a new plugin and after that we can add this as additional feature to `imagespacingbox` plugin.

ck.sh
```diff
@@ -61,6 +61,9 @@ case "$COMMAND" in
 				# Copy custom skin to ckeditor-dev
 				cp -r skins/moono-lexicon ckeditor-dev/skins/moono-lexicon
 
+				# Copy lang files from image plugin to imagespacingbox
+				cp -r ckeditor-dev/plugins/image/lang plugins/imagespacingbox/
```
carloslancha, line 65: Do we want to hardcode this in the build process?

mateomustapic: By following Julien's comment here https://github.com/liferay/liferay-ckeditor/pull/126#discussion_r480021566 , yes

julien: Hey @mateomustapic that comment was a suggestion, if it can be improved than lets improve it.

For the moment there's nothing "generic" that copies stuff (like directories in this case) from existing plugins to our custom plugins directories - but it would just be a matter of adding a function.

That said we don't always want to copy `lang` directories from a specific plugin to a plugin's directory.

As for 

> Do we want to hardcode this in the build process?

We're already hard coding the same kind of stuff on line 62 - and much more in that script.

carloslancha: > We're already hard coding the same kind of stuff on line 62 - and much more in that script.

Yep, I know (and I don't like it xD). Everything hardcoded is related to `moono-lexicon` skin 'cause we don't expect to have more skins (but I still don't like it). I'll take a look to see if we can improve that, I think we're adding too much specific complexity to the build process.

For this one since there's no other plugin languages that need to be copied we can leave it as it is, just wanted to rise up my concern about starting to hardcore everything in the build script ;)

wincent: > my concern about starting to hardcore everything

hardcore

carloslancha: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia1.giphy.com%2Fmedia%2F4WEVDuEbYosIbjnTBG%2Fgiphy.gif&token=hardcore" data-gitmeme-token="hardcore"></a>

julien: @carloslancha yes it's not perfect, but it works, I agree that in the long term if we write more and more plugins we'll need something that handles this "automagically"

We can think of what should be done for each plugin (it might seem like an overkill) - that would probably mean writing something like we did for the plugin's dependencies in the `codemirror` plugin, maybe a `plugin.json` file (instead of our good old `deps.json`)

```json
{
  "name": "someplugin",
  "dependencies": [
      "foo": "path/to/foo",
      "bar": "path/o/bar"
   ],
   "languages": [
        "path/to/ckeditor-dev/plugins/someplugin/lang",

    ],
}
```

The we can have a JS script (because you don't like bash right :joy: ) parse the json and do it's magic.

But I think we should be careful about this

wincent: A great philosopher once said:

> Dumb is the new clever.

Or maybe it was:

> Clever is the new dumb.

Actually, now that I think about it, it wasn't a great philosopher; it was just some guy I used to work with and ever since I heard it I have been repeating it whenever I get the chance. 😂

Playing on Knuth's famous "Premature optimization is the root of all evil", ~another great philosopher~ some person on Hacker News once quipped:

> Premature Abstraction is the root of all evil.

To which [another commenter replied](https://news.ycombinator.com/item?id=9442140):

> Premature optimization, that's like a fart. Premature abstraction is like taking a dump on another developer's desk.

All of which is to say, beware of over-engineering this.

julien: I agree, that's why I said that it might seem like an overkill ...

carloslancha: Quoting myself 🤷 :
> For this one since there's no other plugin languages that need to be copied we can leave it as it is

wincent: friends

mateomustapic: In conclusion, as I understand, this plugin should be done now and doesn't require any more work?

carloslancha: Could you please rebase with master and push? Don't know why is not getting last merged changed to check the format on plugins

mateomustapic: Rebased and pushed, please check now.

carloslancha: I think this is not being build with the `--no-ie-checks` but we can take a look on master after merging this one. LGTM

carloslancha:

ck.sh
carloslancha: Hey @mateomustapic I'm rebasing this pr since have conflicts with recent master due to the changes introduced by https://github.com/liferay/liferay-ckeditor/pull/128

While doing it I realize that we are copying the language files directly in our working plugins folder, that will result in files that we will need to commit or discard, but I think what we want to do is copy them to the building folder (`ckeditor-dev` here, `temp` after rebasing, so

```
cp -r ckeditor-dev/plugins/image/lang plugins/imagespacingbox/
```
will become
```
cd ckeditor-dev
cp -r plugins/images/lang plugins/imagespacingbox
```
and after rebase
```
cd temp
cp -r plugins/images/lang plugins/imagespacingbox
```
Does this make sense to  you?

mateomustapic: Hi, yes I see this generates language files to both `ckeditor-dev/plugins` and working `plugins` directory.
Will you make this change to the script or should I?

carloslancha: I did it already while rebasing with master and pushed to this branch. Please, test the pr with the latest changes and if everything is working as expected, since @julien approved it, feel free to merge it. Thx.

PR #128, opened 2020-09-02
refactor: move plugins skins to plugins folder

carloslancha:
Simplifying here how plugins skins are being included into the build process and the build process itself.

Now we create a temporal directory from which we build CKEditor, so we don't need to worry about changes happening in ckeditor-dev that could interfere with patches or with not properly removed files


ck.sh
```diff
@@ -58,36 +58,35 @@ case "$COMMAND" in
...
-				pluginsWithSkins=$(find ckeditor-dev/plugins -maxdepth 2 -mindepth 2 -type d -name skins)
-				for plugin in $pluginsWithSkins; do
-					pluginDir=$(dirname "$plugin")
-					pluginName=$(basename "$pluginDir")
-					pluginSkinPath="skins/moono-lexicon/plugins/$pluginName/skins/moono-lexicon"
-					if [[ -d "$pluginSkinPath" ]]; then
-						cp -r "$pluginSkinPath" "$pluginDir/skins"
-					fi
-				done
+				# Clean and make a temporal copy of ckeditor-dev to work with
```
wincent, line 61: "temporal" means "relating to time"; I think you want "temporary" here; same on 107.

ck.sh
```diff
@@ -96,16 +95,17 @@ case "$COMMAND" in
...
 				# Remove old build files.
 				rm -rf ../ckeditor/*
 
 				# Replace with new build files.
 				cp -r dev/builder/release/ckeditor/* ../ckeditor/
 
 				cd ..
+
+				# Remove temporal directory
+				rm -rf temp
```
wincent, line 107: In case we don't get to this line, might want to add `/temp` to `.gitignore` file so that it doesn't clutter `git status` output.

ck.sh
```diff
@@ -58,36 +58,35 @@ case "$COMMAND" in
...
-					pluginDir=$(dirname "$plugin")
-					pluginName=$(basename "$pluginDir")
-					pluginSkinPath="skins/moono-lexicon/plugins/$pluginName/skins/moono-lexicon"
-					if [[ -d "$pluginSkinPath" ]]; then
-						cp -r "$pluginSkinPath" "$pluginDir/skins"
-					fi
-				done
+				# Clean and make a temporal copy of ckeditor-dev to work with
+				rm -rf temp
+				cp -r ckeditor-dev temp
```
wincent, line 63: I feel a bit queasy about copying an entire Git submodule (in practice, at least, the `.git` directory is not a real directory, but a "gitlink" file, so at least we're not copying that). Right now, it doesn't have anything that could be "tricky" (eg. symlinks), but it might in the future. We're talking about > 100MB of stuff.

wincent (commented):
Seems like a fair idea, but I'll wait for ~The Bash Guy~ @julien to give his blessing.

ck.sh
carloslancha: <a href="https://gitme.me/image?url=https%3A%2F%2Fmedia0.giphy.com%2Fmedia%2FweNyGjdcQ3d9S%2Fgiphy.gif&token=temporal" data-gitmeme-token="temporal"></a>

ck.sh
carloslancha: Sure

ck.sh
julien: Yes, sounds like a good approach for now.

support/iconsClassesGenerator.js
```diff
@@ -99,9 +98,6 @@ const getCSS = (svgData, cKEditorIcon, direction) => {
 let iconsCSSContent = '';
 
 for (const [cKEditorIcon, clayIcon] of Object.entries(iconsConfig.icons)) {
-	let directionClass = '';
-	let svgDataList = [];
-
```
julien, line 104: :+1

carloslancha:
Updated!

julien (approved):

PR #131, opened 2020-09-15
fix: display formatted text in Firefox Preview Source dialog

mateomustapic:
This PR is a fix for [LPS-120763](https://issues.liferay.com/browse/LPS-120763) - `"CKEditor - Formatted text doesn't display in Preview Source dialog on Firefox"` bug in `codemirror` plugin.

As described in the ticket, the issue happened when editing text and opening Source dialog in Firefox.
The added text would't appear in the code preview area.
This cause of this issue is that in Firefox, the frame's content is not recognised if there is no initial content set. 
To solve this we need to trigger a `page load` and then set `innerHtml` to `iFrame` preview.

--------------
To test this:
1. Build CKEditor with this change
2. Access to portal on Firefox browser.
3. Go to Web Content.
4. Add a content.
5. In content field, click Source button in ckeditor toolbar to go to source view.
6. Type "test" or <p>test</p>
7. Click the Expand icon next to the Source button to preview source.

Old behaviour
Firefox-web content source preview

New behaviour
after

### DIFF
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -44,10 +44,7 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				.getElement();
 			preview.setSize('width', defaultWidth);
 
-			var iframe = this._createContentIframe(preview);
-			var iframeBody = iframe.$.contentDocument.body;
-
-			iframeBody.innerHTML = codeMirrorEditor.getValue();
+			this._createContentIframe(preview);
 
 			codeMirrorEditor.on(
 				'change',
@@ -87,6 +84,57 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				tabPanel.getSize('height') - (padding.bottom + padding.top);
 
 			var iframe = new CKEDITOR.dom.element('iframe');
+
+			function handleStyles() {
+				var iframeDocument = iframe.$.contentDocument;
+
+				var iframeHead = iframeDocument.head;
+
+				var contentsCss = editor.config.contentsCss;
+
+				if (Array.isArray(contentsCss)) {
+					contentsCss.forEach(function (url) {
+						var link = iframeDocument.createElement('link');
+						link.setAttribute('href', url);
+						link.setAttribute('rel', 'stylesheet');
+
+						iframeHead.appendChild(link);
+					});
+				} else {
+					var link = iframeDocument.createElement('link');
+					link.setAttribute('href', contentsCss);
+					link.setAttribute('rel', 'stylesheet');
+
+					iframeHead.appendChild(link);
+				}
+
+				var direction = editor.config.contentsLangDirection;
+
+				var iframeHtml = iframeDocument.documentElement;
+				iframeHtml.setAttribute('dir', direction);
+				iframeHtml.setAttribute('lang', editor.config.defaultLanguage);
+
+				var iframeBody = iframeDocument.body;
+				iframeBody.classList.add('cke_editable');
+				iframeBody.classList.add('cke_editable_themed');
+				iframeBody.classList.add('cke_contents_' + direction);
+
+				iframeBody.setAttribute('contenteditable', false);
+				iframeBody.setAttribute('spellcheck', false);
+
+				iframeBody.style.background = '#fff';
+			}
+
+			var data = this.codeMirrorEditor.getValue();
+
+			iframe.on('load', function () {
+				var iframeBody = iframe.$.contentDocument.body;
+
+				iframeBody.innerHTML = data;
+
+				handleStyles();
+			});
+
 			parentElement.append(iframe);
 
 			iframe.setAttributes({
@@ -98,43 +146,6 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				height: height + 'px',
 				width: '99%',
 			});
-
-			var iframeDocument = iframe.$.contentDocument;
-			var iframeHead = iframeDocument.head;
-
-			var contentsCss = editor.config.contentsCss;
-
-			if (Array.isArray(contentsCss)) {
-				contentsCss.forEach(function (url) {
-					var link = iframeDocument.createElement('link');
-					link.setAttribute('href', url);
-					link.setAttribute('rel', 'stylesheet');
-
-					iframeHead.appendChild(link);
-				});
-			} else {
-				var link = iframeDocument.createElement('link');
-				link.setAttribute('href', contentsCss);
-				link.setAttribute('rel', 'stylesheet');
-
-				iframeHead.appendChild(link);
-			}
-
-			var direction = editor.config.contentsLangDirection;
-
-			var iframeHtml = iframeDocument.documentElement;
-			iframeHtml.setAttribute('dir', direction);
-			iframeHtml.setAttribute('lang', editor.config.defaultLanguage);
-
-			var iframeBody = iframeDocument.body;
-			iframeBody.classList.add('cke_editable');
-			iframeBody.classList.add('cke_editable_themed');
-			iframeBody.classList.add('cke_contents_' + direction);
-
-			iframeBody.setAttribute('contenteditable', false);
-			iframeBody.setAttribute('spellcheck', false);
-
-			iframeBody.style.background = '#fff';
 
 			return iframe;
 		},
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -84,56 +84,7 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				tabPanel.getSize('height') - (padding.bottom + padding.top);
 
 			var iframe = new CKEDITOR.dom.element('iframe');
-
-			function handleStyles() {
-				var iframeDocument = iframe.$.contentDocument;
-
-				var iframeHead = iframeDocument.head;
-
-				var contentsCss = editor.config.contentsCss;
-
-				if (Array.isArray(contentsCss)) {
-					contentsCss.forEach(function (url) {
-						var link = iframeDocument.createElement('link');
-						link.setAttribute('href', url);
-						link.setAttribute('rel', 'stylesheet');
-
-						iframeHead.appendChild(link);
-					});
-				} else {
-					var link = iframeDocument.createElement('link');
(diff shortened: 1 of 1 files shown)


carloslancha:
Hey @mateomustapic !

This fix feels weird to me. `document.open()` is meant to be used to prepared the document to write on it (`document.write`), and causes some side effects like remove all nodes and listeners attached to it.

Doing a quick test looks like if, with the original code, you add a breakpoint after the iframe creation and before setting the `innerHTML` value, everything is working as expected, so maybe we're facing some kind of career condition? Something related with the iframe load event?

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -199,9 +207,7 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 			var editor = this.getParentEditor();
 			var data = editor.getData();
 
-			if (codeMirrorEditor && codeMirrorEditor.getValue() !== data) {
```
carloslancha, line 202: What I understand we're doing here is checking if codeMirror exists and update the source editor content only if its current value is different from the editor one. Why we need to remove this?

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -43,7 +43,15 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
...
-			var iframeBody = iframe.$.contentDocument.body;
+
+			var iframeWindow = iframe.$.contentWindow;
+
+			var iframeBody = iframeWindow.document.body;
+
+			var iframeDocument = iframeWindow.document;
+
+			iframeDocument.open();
+			iframeDocument.close();
```
julien, line 54: @mateomustapic what @carloslancha mentioned [here](https://github.com/liferay/liferay-ckeditor/pull/131#issuecomment-692711583) is also what I had in mind and for me it makes sense.
I tried with 

```js
iframe.$.contentWindow.addEventListener('load', function() {
  var iframeBody = iframe.$.contentDocument.body;
  iframeBody.innerHTML = codeMirrorEditor.getValue();
});
``` 
and that works for me

julien (changes requested):

plugins/codemirror/dialogs/codemirrordialog.js
mateomustapic: hey @julien, Thanks for this. 
Yesterday at that moment, I didn't have better solution than the one I provided.
Having in mind what Carlos mentioned earlier, I will modify and push the change with this code snippet you provided.

mateomustapic:
> Hey @mateomustapic !
> 
> This fix feels weird to me. `document.open()` is meant to be used to prepare the document to write on it (`document.write`), and causes some side effects like remove all nodes and listeners attached to it.
> 
> Doing a quick test looks like if, with the original code, you add a breakpoint after the iframe creation and before setting the `innerHTML` value, everything is working as expected, so maybe we're facing some kind of career condition? Something related with the iframe load event?



> Hey @mateomustapic !
> 
> This fix feels weird to me. `document.open()` is meant to be used to prepare the document to write on it (`document.write`), and causes some side effects like remove all nodes and listeners attached to it.
> 
> Doing a quick test looks like if, with the original code, you add a breakpoint after the iframe creation and before setting the `innerHTML` value, everything is working as expected, so maybe we're facing some kind of career condition? Something related with the iframe load event?

Hey @carloslancha  I am aware that this fix was a bit weird, but it was the only solution I could at the time. 
I am resending now.

plugins/codemirror/dialogs/codemirrordialog.js
julien: @mateomustapic no problem, that's why we do reviews, just make sure that what I suggested works

julien:
Hey @mateomustapic,

This works in Firefox

firefox

But now it's not working in Chrome

chrome

It also seems that the styles in Firefox are not loaded properly, so if it doesn't seem to hard we could fix that as well, otherwise, we'll have to create an issue

mateomustapic:
> Hey @mateomustapic,
> 
> This works in Firefox
> 
> firefox
> 
> But now it's not working in Chrome
> 
> chrome
> 
> It also seems that the styles in Firefox are not loaded properly, so if it doesn't seem to hard we could fix that as well, otherwise, we'll have to create an issue

hey @julien 
If you are referring to styles described in this issue https://issues.liferay.com/browse/LPS-120758, I have prepared a fix for it.
I am now going to working on fixing the issue of not working in Chrome.

julien:
@mateomustapic I was thinking about the fonts in the preview, but we can see that later.

mateomustapic:
Hey @julien 

I have pushed the change where this fix is working both in Chrome and in Firefox.

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -43,9 +43,17 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 			preview.setSize('width', defaultWidth);
 
 			var iframe = this._createContentIframe(preview);
+
 			var iframeBody = iframe.$.contentDocument.body;
 
-			iframeBody.innerHTML = codeMirrorEditor.getValue();
+			if (CKEDITOR.env.gecko) {
```
carloslancha, line 49: Hey! The fact that we need to add this kind of condition here makes me think if this is the actual problem or if it's lying a little deeper.

The `load` event should be fired in the same way in both browsers, so maybe we're adding the listener too late?

mateomustapic: This condition is added because `load` event in `iframe` is handled differently in Chrome/Safari and Firefox.
The issue is when appending iframe to document.
Due to multiple use cases I found while researching this, we cannot fire this event the same way to make it work in all browsers.

carloslancha: Did you try adding the listener to the iframe's `onload` event before appending it to the document?

julien: @carloslancha is right we can handle this the same way both in Chrome or Firefox (and don't forget to check IE11 please), I checked before comenting on #133

julien:
@mateomustapic this fixes the issue in Firefox but it looks like you forgot to update everything.
You're setting innerHTML [here](https://github.com/liferay/liferay-ckeditor/blob/16fb031cb47b2f73b1f562e87321cd647eabfc94/plugins/codemirror/dialogs/codemirrordialog.js#L48-L50) and [here](https://github.com/liferay/liferay-ckeditor/blob/16fb031cb47b2f73b1f562e87321cd647eabfc94/plugins/codemirror/dialogs/codemirrordialog.js#L96). Are you sure both are needed?

Given that the styles aren't present in Firefox, I think we could have taken care of this in the same PR, but that's just my opinion - I'll wait for @carloslancha to share his thoughts on that

mateomustapic:
> @mateomustapic this fixes the issue in Firefox but it looks like you forgot to update everything.
> You're setting innerHTML [here](https://github.com/liferay/liferay-ckeditor/blob/16fb031cb47b2f73b1f562e87321cd647eabfc94/plugins/codemirror/dialogs/codemirrordialog.js#L48-L50) and [here](https://github.com/liferay/liferay-ckeditor/blob/16fb031cb47b2f73b1f562e87321cd647eabfc94/plugins/codemirror/dialogs/codemirrordialog.js#L96). Are you sure both are needed?
> 
> Given that the styles aren't present in Firefox, I think we could have taken care of this in the same PR, but that's just my opinion - I'll wait for @carloslancha to share his thoughts on that

yes, you are right. I will now push the code without setting the "first" `innerHTML`.

mateomustapic:
I pushed the changes containing previously discussed fix for `Firefox` and moved styles for `Firefox` in `load` callback. 
This was done because in Firefox stylesheets are not added to the iframe's head, so styles for `Preview` dialog were not loaded as in `Chrome`.

julien:
@mateomustapic I checked and it is fixed for Firefox.



I was wondering, if we shouldn't but all that code in it's own "named" function instead of an anonymous one, it can always be done later but since this is the type of thing we're likely to forget or postpone, I'd suggest doing it now. 

Another thing I noticed is that when we do a build (without setting `DEBUG=1`) there is a warning concerning the [`class`](https://github.com/liferay/liferay-ckeditor/blob/c3cc4a397829c9713c4e54e1be1bf53ad89145a9/plugins/codemirror/dialogs/codemirrordialog.js#L135) keyword that we could probably get rid of by surrounding it in quotes.

mateomustapic:
I agree with you @julien on this. 
It is better to have a named function for the readability.
Regarding second comment, I haven’t noticed it since I build editor with ‘DEBUG=1’, but I can do that too in the next iteration.

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -87,6 +84,51 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				tabPanel.getSize('height') - (padding.bottom + padding.top);
 
 			var iframe = new CKEDITOR.dom.element('iframe');
+
+			var data = this.codeMirrorEditor.getValue();
+
+			iframe.on('load', function () {
```
carloslancha, line 90: As @julien pointed out we should extract this to a named function. I'm also wondering why we need this code to be executed twice, since we're running the exact same stuff starting at line 144 and again on this `load` event. Is this necessary?

carloslancha:
> Regarding second comment, I haven’t noticed it since I build editor with ‘DEBUG=1’, but I can do that too in the next iteration.

As a good practice I recommend to test the build not in debug mode before sending / updating a pr. This way we'll be sure we're not introducing things CKBuilder will be complaining about that the one who will prepare the release will have to fix.

plugins/codemirror/dialogs/codemirrordialog.js
julien: no it's not necessary Carlos, I checked the diff quickly and assumed the code had been moved, not duplicated - thanks for pointing this out.

mateomustapic: hey @carloslancha , it is not necessary. 
It was Friday so I wanted to finish it quickly, so it seems I forgot to remove the code I moved :D.

carloslancha (approved):

julien (approved):

julien:
I also got rid of the warning that was emitted when building - due to the use of the `class` keyword
Waiting for CI to go green and I'll merge

PR #132, opened 2020-09-17
fix: enable line wrapping in codemirror plugin

julien:
This solves the issue described in [LPS-120594](https://issues.liferay.com/browse/LPS-120549), where
occasionally the editor's source mode (now handled by the `codemirror`
plugin) grows depending on the text that is in the editor.

Here's a screenshot of what it looked like before this change:

And here's a screenshot of what it looks like after this change:

**Test plan**:

- Fetch this branch, and build `liferay-ckeditor` with `./ck.sh build`
- Use the previously built `liferay-ckeditor` in DXP. ([Instructions](https://github.com/liferay/liferay-ckeditor#testing-in-liferay-portal))
- In DXP, from the global menu, navigate to "Instance Settings > Email" and once the page is loaded, click on the "Source" button 

Another thought I had, is that if we were using ES6 in CKEDITOR plugins, 
we could possibly add the CodeMirror options in a separate file to make sure
both editor (the one in `codemirror` plugin and the one in `codemirrordialog`) 
always have the same options - but I'm not ready yet (and who knows
what could happen if we decided to build CKEDITOR with ES6)

### DIFF
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -33,7 +33,14 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 
 			var defaultWidth = size.width / 2 - padding;
 
-			codeMirrorEditor.setSize(defaultWidth, null);
+			var tabPanel = this._getTabPanel();
+			var tabPanelPadding = this._getTabPanelPadding();
+
+			var defaultHeight =
+				tabPanel.getSize('height') -
+				(tabPanelPadding.bottom + tabPanelPadding.top);
+
+			codeMirrorEditor.setSize(defaultWidth, defaultHeight);
 
 			var editor = dialog.getParentEditor();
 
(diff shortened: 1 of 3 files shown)


julien:
p.s. Please don't merge this yet, as I'm fixing some similar issue in the dialog as commented in #133

julien:
In [836ffcb](https://github.com/liferay/liferay-ckeditor/pull/132/commits/836ffcb26e1c359eef0e8700fbc997fd4f885362) I also fixed the dialog's height which looks a bit better, but I still need to take care of "autowrapping" the text in there too

julien:
[`7acbcee`](https://github.com/liferay/liferay-ckeditor/pull/132/commits/7acbcee176237c1d8937ef73e1d9d327d126c0f8) adds text wrapping in the `codemirrordialog` too

julien:
Now resolving conflicts due to the changes in #131

julien:
I still need to remove duplicated code here so I'll add another commit

PR #133, opened 2020-09-17
fix: modify calculation for dialog defaultWidth which caused visual bug

mateomustapic:
This PR fixes an issue described in https://issues.liferay.com/browse/LPS-120758 ,visual bug in `Chrome` (scrollbar) and in `Firefox` (whitespace) in Source dialog box.
This issue was caused by `CodeMirror` plugin line number width not included in dialog size calculation.

To test this fix:
1. Build CKEditor with this change
2. Access to portal.
3. Go to Web Content.
4. Add a content.
5. In content field, click Source button in ckeditor toolbar to go to source view.
6. Type any text
7. Click the Expand icon next to the Source button to preview source dialog.

Before Chrome
beforechrome

Before Firefox
beforefirefox

After Chrome
afterchrome

After Firefox
afterfirefox

### DIFF
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -29,7 +29,9 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				}
 			));
 
-			var defaultWidth = size.width * 0.5 - 10;
+			var padding = 40;
+
+			var defaultWidth = size.width / 2 - padding;
 
 			codeMirrorEditor.setSize(defaultWidth, null);
 


plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -29,7 +29,9 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 				}
 			));
 
-			var defaultWidth = size.width * 0.5 - 10;
+			var padding = 40;
+
+			var defaultWidth = size.width * 0.5 - padding;
```
carloslancha, line 34: Is this `padding` var being used elsewhere?  If not I'd avoid creating the variable

carloslancha (commented):
Just left a nit. @julien will know better the impact of this change :)

plugins/codemirror/dialogs/codemirrordialog.js
julien: you're right, it's not being used anywhere, but I'm guessing @mateomustapic added this to avoid a "magic number"

julien:
gonna test this right now :partying_face:

plugins/codemirror/dialogs/codemirrordialog.js
carloslancha: Is not 0.5 already a magic number? 😛

wincent: Well, it translates pretty obviously to "half" so not all that magical, unlike `padding` which is totally arbitrary. You probably wouldn't want to write:

```js
var padding = 40;

var half = 0.5;

var defaultWidth = size.width * half - padding;
```

julien: > Is not 0.5 already a magic number? stuck_out_tongue

yeah, well you make a point;  but I always read that as "whatever / 2"

x / 2 == x * 0.5

julien:
@mateomustapic so I tested this and here's what I think (with screenshots):

The bug seems to be gone in Chrome:



and also seems fixed when you have the devtools open or if the window has a smaller size:



Same thing, in Firefox:




And in Firefox with the devtools open:

firefox02

But while we're looking at it, I also would consider fixing this, because it seems related (I'm not saying it's mandatory but if it's not to hard to fix, why not)

plugins/codemirror/dialogs/codemirrordialog.js
julien: Sorry I hadn't seen @wincent 's comment before adding mine: now that I've read it, I agree on avoiding **those** magic numbers everywhere

julien:
@mateomustapic concerning the editor's height in the dialog, I'm going to to that in #132
Just make sure this works fine in IE11 as well please. Thanks

mateomustapic:
> @mateomustapic concerning the editor's height in the dialog, I'm going to to that in #132
> Just make sure this works fine in IE11 as well please. Thanks

hey @julien 

Just tested this in IE, and works as in other browsers.
Before
ie_before

After
ie_witdth_test

julien:
Ok thanks for confirming @mateomustapic, I still going to give it a quick test run in DXP because who knows.

Could you tell me what you think of @wincent's [comment](https://github.com/liferay/liferay-ckeditor/pull/133#discussion_r490259709)?

I personally think it's better to avoid those magic numbers, so I'd be OK to do it both for `half` (0.5) and `padding` (40)

 @carloslancha does that seem reasonable to you as well?

mateomustapic:
> Ok thanks for confirming @mateomustapic, I'm still going to give it a quick test run in DXP because who knows.
> 
> Could you tell me what you think of @wincent's [comment](https://github.com/liferay/liferay-ckeditor/pull/133#discussion_r490259709)?
> 
> I personally think it's better to avoid those magic numbers, so I'd be OK to do it both for `half` (0.5) and `padding` (40)
> 
> @carloslancha does that seem reasonable to you as well?

@julien As I can see, @wincent says  "_Well, it translates pretty obviously to "half" so not all that magical, unlike padding which is totally arbitrary. You probably wouldn't want to write .... var defaultWidth = size.width * half - padding;" , meaning adding of `padding` variable is ok, but not adding of `half` variable, where 0.5 value is self-explanatory.

julien:
I need to wear glasses

jbalsas:
Why don't simply use `foo / 2`? 

That's more obviously a "halving" operation than "*.5" :)

julien:
@jbalsas that's also true - even more obvious

mateomustapic:
yeah, it's more readable to  divide by 2 than multiply 0.5.

julien:
The change looks good to me - so I'm merging this one

Issue #134, opened 2020-09-25
colorbutton removes scrollbar when used in Alloyeditor


fortunatomaldonado:
https://issues.liferay.com/browse/LPS-121472

When Text Color is added to AlloyEditor, [`onblock`](https://github.com/ckeditor/ckeditor4/blob/major/plugins/colorbutton/plugin.js#L193) is called and removes a scrollbar.

It looks like this behavior can be fixed with a condition so that it will not change the intended behavior of CKEditor and fix the issue in AlloyEditor.

Let me know if there are any comments or questions about this.
Thank you.

wincent:
The LPS sounds like a legit bug. Can you put together a PR for us to evaluate?

fortunatomaldonado:
This issue is due to a third party plugin and it is up to the customer to address this. We provided assistance on where to address this issue [here.](https://github.com/liferay/alloy-editor/issues/749#issuecomment-306788867)

PR #135, opened 2020-09-29
fix: add condition for buttoncolor so scrollbar will not be removed

fortunatomaldonado:
https://github.com/liferay/liferay-ckeditor/issues/134
https://issues.liferay.com/browse/LPS-121472

A scrollbar will disappear whenever AlloyEditor contains the Text Color button and it is being used.

I found that CKEditor uses this behavior [`onBlock` ](https://github.com/ckeditor/ckeditor4/blob/major/plugins/colorbutton/plugin.js#L193) for its text color block. I added a condition to check for the class `cke_ltr` which is used only in CKEditor for the Text Color block and not in AlloyEditor. This will prevent any unnecesary changes to CKEditor and will not remove the scrollbar when it is used in AlloyEditor.

Please let me know if there are any questions, comments, or if I need to change any process that I did.
Thank you.

### DIFF
--- a/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
+++ b/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
@@ -1,4 +1,4 @@
-From bb8157502a44ede7c8b1806ef63fcdf3554e8089 Mon Sep 17 00:00:00 2001
+From 113849fd54d137f21f488b0ffe06642e93d3aa6f Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Tue, 14 May 2019 10:47:25 +0200
 Subject: [PATCH] LPS-89596 Cannot Drag Image from top content line in IE11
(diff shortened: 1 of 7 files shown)


carloslancha:
Hey @fortunatomaldonado thx for the pr!

Could you please remove the autogenerated chore commit `update ckeditor build artifacts`? We generate them on the release process so we don't need it here.

Also the Semantic PR checker is complaining about the title of the pr, could you please fix it and add a title that describes the fix?

I'll try to take a look and test the pr later. Thx again!

fortunatomaldonado:
Hello @carloslancha,

Thank you for taking a look at this. I fixed the commit and the title. Please let me know if there is anything else I need to do.

Thank you again!

fortunatomaldonado:
Hello @carloslancha, 
Can you provide me any update on this PR? 
Thank you.

carloslancha:
Hey @fortunatomaldonado !

I took a look at it. As you discovered, the issue is caused due to different panels behavior in CKEditor and AlloyEditor. CKEditor renders an iframe with the panel inside while AlloyEditor doesn't renders an iframe, so when it gets the body of the document `block.element.getDocument().getBody()` CKEditor gets the iframe body but AlloyEditor gets the content body.

The thing is that this is a CKEditor plugin and it's working correctly for CKEditor, so introducing changes here to fix stuff in AlloyEditor feels like not the best solution.

It would be reasonable to add some fixes in [alloy-editor](https://github.com/liferay/alloy-editor). Maybe a good starting point could be [panel-menu-button](https://github.com/liferay/alloy-editor/blob/master/src/components/uibridge/panel-menu-button.jsx)

Feel free to reach us via pr or in #t-frontend-infra slack channel if you need further help.

Thx!!

Issue #138, opened 2020-11-13
CodeMirror fails to load when "Expose Global" is enabled


kevhlee:
In Liferay DXP, the "Source mode" in CKEditor no longer functions when "Expose Global" is enabled. This issue is that the CodeMirror plugin does not properly load. Please see [PTR-2064](https://issues.liferay.com/browse/PTR-2064) for more context.

![JS_Error](https://user-images.githubusercontent.com/21070577/99101703-d0e3ae80-2591-11eb-8542-9aaca5329dab.png)

In `liferay-ckeditor/plugins/codemirror/vendors/vendors.js`, we have the following AMD-related code (unminified for readability):

```javascript
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
  typeof define === 'function' && define.amd ? define(factory) :
  (global = global || self, global.CodeMirror = factory());
}(this, (function () { 'use strict';
```

One solution would be adding the following to the logic:

```javascript
(function (global, factory) {
  typeof exports === 'object' && typeof module !== 'undefined' ? module.exports = factory() :
  false && typeof define === 'function' && define.amd ? define(factory) :
  (global = global || self, global.CodeMirror = factory());
}(this, (function () { 'use strict';
```

By adding `false` to the condition, we would prevent CodeMirror from being loaded using AMD. However, since `liferay-ckeditor/plugins/codemirror/vendors/vendors.js` is auto-generated, we would need to ensure that the change does not get overwritten. I'm not entirely sure how to apply this change, which is why I have not created a PR yet.

If this is not the ideal solution, please feel free to share other solutions. Let me know if you have any questions/thoughts.

cc: @julien, @carloslancha, @wincent

julien:
Fixed in #141

PR #139, opened 2020-11-17
fix: prevent loading CodeMirror via AMD

kevhlee:
**Issue:** #138 

**Relevant Tickets:**
<https://issues.liferay.com/browse/PTR-2064>
<https://issues.liferay.com/browse/LPS-123222>

**Notes:**
This is simply a proof-of-concept. The `vendors.js` file was automatically generated by this [commit](https://github.com/liferay/liferay-ckeditor/commit/696512d26acb7ae72326768249ad85d3393a16e4) after running `ck.sh`, so we need to make sure the change is not overwritten by `ck.sh`, as mentioned in #138. Also, as a note, [LPS-123222](https://issues.liferay.com/browse/LPS-123222) describes an additional issue when "Expose Global" is on.


carloslancha:

kevhlee:
Hi @carloslancha,

Any updates on this issue/fix?

julien:
Hey @kevhlee,


Thanks for reporting this. 
Unfortunately, we don't own the codemirror source code so we can't just modify it. 

This issue looks like it's more of a bug in our AMD Loader than the plugin which is "transpiled" to UMD.

jbalsas:
Hey @julien 

> Unfortunately, we don't own the codemirror source code so we can't just modify it.

We're bringing the code over, so we should be able to patch it when it comes

> This issue looks like it's more of a bug in our AMD Loader than the plugin which is "transpiled" to UMD.

It's not. The UMD wrapper that codemirror libs has is a standard wrapper. In presence of `window.require && window.require.amd` the wrapper will act as AMD rather than a global library. [We expect `Codemirror` to always be global](https://github.com/liferay/liferay-ckeditor/blob/master/plugins/codemirror/plugin.js#L18) so we should either take a non-umd'd version or make sure the library we use always exposes `Codemirror` globallly by negating the AMD clause.

julien:
> We're bringing the code over, so we should be able to patch it when it comes

That's true.

About the AMD loader, my bad and thanks for the explanation.

julien:
Hey @kevhlee,

Just so you know we're working on this, I fetched these changes, built (liferay-)ckeditor in debug mode and deployed it in dxp but I'm still seeing errors:

2020-11-24-145738_754x439_scrot

In any case, I'm going to look out what options we have to build `CodeMirror` as a global.

julien:
So (I'm replying to myself but you can all follow along :joy: ) that's happening because there [are](https://github.com/liferay/liferay-ckeditor/blob/83bac309d50132201ee1a63c6ccb5c74c66f4102/ckeditor/plugins/codemirror/vendors/vendors.js#L1) [several](https://github.com/liferay/liferay-ckeditor/blob/83bac309d50132201ee1a63c6ccb5c74c66f4102/ckeditor/plugins/codemirror/vendors/vendors.js#L329) [calls](https://github.com/liferay/liferay-ckeditor/blob/83bac309d50132201ee1a63c6ccb5c74c66f4102/ckeditor/plugins/codemirror/vendors/vendors.js#L424) to `define(` in that `vendor.js` file - if we apply the same "fix" on those lines it does fix the issue when **Expose Global** is checked for the AMD Loader.

The question now is: how do we fix this correctly?

I've thought of the following solutions:

- Keep everything as is and use `sed` or `awk` to modify those lines (could work but seems a bit like a hack).

- Add CodeMirror as a submodule and patch on top (seems cleaner - but also going to increase the complexity of the build)

I'll keep on thinking, there must be a simple way

wincent:
> * Add CodeMirror as a submodule and patch on top (seems cleaner - but also going to increase the complexity of the build)

Isn't the code already in the ckeditor submodule in this repo, specifically [here](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/samples/toolbarconfigurator/lib/codemirror/codemirror.js)?

If so, just patch it like you would anything else in the submodule.

julien:
@wincent thanks I hadn't seen that.

The thing is that we use CodeMirror's `xml` mode (for syntax highlighting) which might not be included (I have to check) and another dependency called `js-beautify` (because by default CKEDITOR puts the HTML on one line in the source view which isn't very nice) but  I'm going to try and use that and see what can be done.

julien:
The easy solution would be to use the default "source" view plugin from CKEditor if you really want to use the "Expose Global" switch with the AMD Loader.

The [`xml` mode](https://github.com/codemirror/CodeMirror/blob/464a66067b8d984c81b8e61ce048b34d7a1054bb/mode/xml/xml.js#L4-L11) that we use it already committed with the "UMD wrapper" function and there's no way to configure that as far as I've seen.

julien:
continued in #141

PR #141, opened 2020-11-25
fix: rename define in plugin dependencies

julien:
This change is related to #139, instead of touching the minified files directly, here we use `@babel` to rename `define` to `__define_disabled__`: if someone can give me some pointers we could probably remove the entire `if`
but for the moment this fixes the issue.

### DIFF
--- a/support/copyPluginDependencies.js
+++ b/support/copyPluginDependencies.js
@@ -1,6 +1,10 @@
 const fs = require('fs');
 const path = require('path');
 
+const parser = require('@babel/parser');
+const traverse = require('@babel/traverse').default;
+const generate = require('@babel/generator').default;
+
 const pluginDir = process.argv[2];
 
 const configFilePath = path.join(pluginDir, 'deps.json');
@@ -23,7 +27,25 @@ for (const [depType, deps] of Object.entries(depsConfig)) {
 		files.forEach((file) => {
 			const filePath = path.join('../node_modules', dep, file);
 
-			fileContent += fs.readFileSync(filePath);
+			if (depType === 'js') {
+				const code = fs.readFileSync(filePath);
+
+				const ast = parser.parse(code.toString());
+
+				traverse(ast, {
+					enter(path) {
+						if (path.node.name === 'define') {
+							path.node.name = '__define_disabled__';
+						}
+					},
+				});
+
+				const output = generate(ast, code);
+
+				fileContent += output.code;
+			} else {
+				fileContent += fs.readFileSync(filePath);
+			}
 		});
 	}
 
--- a/support/copyPluginDependencies.js
+++ b/support/copyPluginDependencies.js
@@ -40,7 +40,7 @@ for (const [depType, deps] of Object.entries(depsConfig)) {
 							path.node.argument.type === 'Identifier' &&
 							path.node.argument.name === 'define'
 						) {
-							path.replaceWithSourceString('__define_disabled__');
+							path.node.argument.name = '__define_disabled__';
 						}
 					},
 				});
--- a/support/copyPluginDependencies.js
+++ b/support/copyPluginDependencies.js
@@ -34,8 +34,13 @@ for (const [depType, deps] of Object.entries(depsConfig)) {
 
 				traverse(ast, {
 					enter(path) {
-						if (path.node.name === 'define') {
-							path.node.name = '__define_disabled__';
+						if (
+							path.node.type === 'UnaryExpression' &&
+							path.node.operator === 'typeof' &&
+							path.node.argument.type === 'Identifier' &&
+							path.node.argument.name === 'define'
+						) {
+							path.replaceWithSourceString('__define_disabled__');
 						}
 					},
 				});
--- a/package.json
+++ b/package.json
@@ -23,5 +23,10 @@
 		"postversion": "npx liferay-js-publish",
 		"format": "prettier --write \"(plugins|skins)/**/*.css\" \"(plugins|support)/**/*.js\" \"*.json\" \"*.md\"",
 		"format:check": "prettier --list-different \"(plugins|skins)/**/*.css\" \"(plugins|support)/**/*.js\" \"*.json\" \"*.md\""
+	},
+	"dependencies": {
+		"@babel/generator": "^7.12.5",
+		"@babel/parser": "^7.12.7",
+		"@babel/traverse": "^7.12.9"
 	}
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,97 @@
 # yarn lockfile v1
 
 
+"@babel/code-frame@^7.10.4":
+  version "7.10.4"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.10.4.tgz#168da1a36e90da68ae8d49c0f1b48c7c6249213a"
+  integrity sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==
+  dependencies:
+    "@babel/highlight" "^7.10.4"
+
+"@babel/generator@^7.12.5":
+  version "7.12.5"
+  resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.12.5.tgz#a2c50de5c8b6d708ab95be5e6053936c1884a4de"
+  integrity sha512-m16TQQJ8hPt7E+OS/XVQg/7U184MLXtvuGbCdA7na61vha+ImkyyNM/9DDA0unYCVZn3ZOhng+qz48/KBOT96A==
+  dependencies:
+    "@babel/types" "^7.12.5"
+    jsesc "^2.5.1"
+    source-map "^0.5.0"
+
+"@babel/helper-function-name@^7.10.4":
+  version "7.10.4"
+  resolved "https://registry.yarnpkg.com/@babel/helper-function-name/-/helper-function-name-7.10.4.tgz#d2d3b20c59ad8c47112fa7d2a94bc09d5ef82f1a"
+  integrity sha512-YdaSyz1n8gY44EmN7x44zBn9zQ1Ry2Y+3GTA+3vH6Mizke1Vw0aWDM66FOYEPw8//qKkmqOckrGgTYa+6sceqQ==
+  dependencies:
+    "@babel/helper-get-function-arity" "^7.10.4"
+    "@babel/template" "^7.10.4"
+    "@babel/types" "^7.10.4"
+
+"@babel/helper-get-function-arity@^7.10.4":
+  version "7.10.4"
+  resolved "https://registry.yarnpkg.com/@babel/helper-get-function-arity/-/helper-get-function-arity-7.10.4.tgz#98c1cbea0e2332f33f9a4661b8ce1505b2c19ba2"
+  integrity sha512-EkN3YDB+SRDgiIUnNgcmiD361ti+AVbL3f3Henf6dqqUyr5dMsorno0lJWJuLhDhkI5sYEpgj6y9kB8AOU1I2A==
+  dependencies:
+    "@babel/types" "^7.10.4"
+
+"@babel/helper-split-export-declaration@^7.11.0":
+  version "7.11.0"
+  resolved "https://registry.yarnpkg.com/@babel/helper-split-export-declaration/-/helper-split-export-declaration-7.11.0.tgz#f8a491244acf6a676158ac42072911ba83ad099f"
+  integrity sha512-74Vejvp6mHkGE+m+k5vHY93FX2cAtrw1zXrZXRlG4l410Nm9PxfEiVTn1PjDPV5SnmieiueY4AFg2xqhNFuuZg==
+  dependencies:
+    "@babel/types" "^7.11.0"
+
+"@babel/helper-validator-identifier@^7.10.4":
+  version "7.10.4"
+  resolved "https://registry.yarnpkg.com/@babel/helper-validator-identifier/-/helper-validator-identifier-7.10.4.tgz#a78c7a7251e01f616512d31b10adcf52ada5e0d2"
+  integrity sha512-3U9y+43hz7ZM+rzG24Qe2mufW5KhvFg/NhnNph+i9mgCtdTCtMJuI1TMkrIUiK7Ix4PYlRF9I5dhqaLYA/ADXw==
+
+"@babel/highlight@^7.10.4":
+  version "7.10.4"
+  resolved "https://registry.yarnpkg.com/@babel/highlight/-/highlight-7.10.4.tgz#7d1bdfd65753538fabe6c38596cdb76d9ac60143"
+  integrity sha512-i6rgnR/YgPEQzZZnbTHHuZdlE8qyoBNalD6F+q4vAFlcMEcqmkoG+mPqJYJCo63qPf74+Y1UZsl3l6f7/RIkmA==
+  dependencies:
+    "@babel/helper-validator-identifier" "^7.10.4"
+    chalk "^2.0.0"
+    js-tokens "^4.0.0"
+
+"@babel/parser@^7.12.7":
+  version "7.12.7"
+  resolved "https://registry.yarnpkg.com/@babel/parser/-/parser-7.12.7.tgz#fee7b39fe809d0e73e5b25eecaf5780ef3d73056"
+  integrity sha512-oWR02Ubp4xTLCAqPRiNIuMVgNO5Aif/xpXtabhzW2HWUD47XJsAB4Zd/Rg30+XeQA3juXigV7hlquOTmwqLiwg==
+
+"@babel/template@^7.10.4":
+  version "7.12.7"
+  resolved "https://registry.yarnpkg.com/@babel/template/-/template-7.12.7.tgz#c817233696018e39fbb6c491d2fb684e05ed43bc"
+  integrity sha512-GkDzmHS6GV7ZeXfJZ0tLRBhZcMcY0/Lnb+eEbXDBfCAcZCjrZKe6p3J4we/D24O9Y8enxWAg1cWwof59yLh2ow==
+  dependencies:
+    "@babel/code-frame" "^7.10.4"
+    "@babel/parser" "^7.12.7"
+    "@babel/types" "^7.12.7"
+
+"@babel/traverse@^7.12.9":
+  version "7.12.9"
+  resolved "https://registry.yarnpkg.com/@babel/traverse/-/traverse-7.12.9.tgz#fad26c972eabbc11350e0b695978de6cc8e8596f"
+  integrity sha512-iX9ajqnLdoU1s1nHt36JDI9KG4k+vmI8WgjK5d+aDTwQbL2fUnzedNedssA645Ede3PM2ma1n8Q4h2ohwXgMXw==
+  dependencies:
+    "@babel/code-frame" "^7.10.4"
+    "@babel/generator" "^7.12.5"
+    "@babel/helper-function-name" "^7.10.4"
+    "@babel/helper-split-export-declaration" "^7.11.0"
+    "@babel/parser" "^7.12.7"
+    "@babel/types" "^7.12.7"
+    debug "^4.1.0"
+    globals "^11.1.0"
+    lodash "^4.17.19"
+
+"@babel/types@^7.10.4", "@babel/types@^7.11.0", "@babel/types@^7.12.5", "@babel/types@^7.12.7":
+  version "7.12.7"
+  resolved "https://registry.yarnpkg.com/@babel/types/-/types-7.12.7.tgz#6039ff1e242640a29452c9ae572162ec9a8f5d13"
+  integrity sha512-MNyI92qZq6jrQkXvtIiykvl4WtoRrVV9MPn+ZfsoEENjiWcBQ3ZSHrkxnJWgWtLX3XXqX5hrSQ+X69wkmesXuQ==
+  dependencies:
+    "@babel/helper-validator-identifier" "^7.10.4"
+    lodash "^4.17.19"
+    to-fast-properties "^2.0.0"
+
 "@clayui/css@^3.14.0":
   version "3.14.0"
   resolved "https://registry.yarnpkg.com/@clayui/css/-/css-3.14.0.tgz#5b9093322b2ca8bfe6c6edea960c8dcc89f5ddc3"
@@ -11,6 +102,13 @@ abbrev@1:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
   integrity sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==
+
+ansi-styles@^3.2.1:
+  version "3.2.1"
+  resolved "https://registry.yarnpkg.com/ansi-styles/-/ansi-styles-3.2.1.tgz#41fbb20243e50b12be0f04b8dedbf07520ce841d"
+  integrity sha512-VT0ZI6kZRdTh8YyJw3SMbYm/u+NqfsAxEpWO0Pf9sq8/e94WxxOpPKx9FR1FlyCtOVDNOQ+8ntlqFxiRc+r5qA==
+  dependencies:
+    color-convert "^1.9.0"
 
 balanced-match@^1.0.0:
   version "1.0.0"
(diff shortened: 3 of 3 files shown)


support/copyPluginDependencies.js
```diff
@@ -23,7 +27,25 @@ for (const [depType, deps] of Object.entries(depsConfig)) {
...
 
-			fileContent += fs.readFileSync(filePath);
+			if (depType === 'js') {
+				const code = fs.readFileSync(filePath);
+
+				const ast = parser.parse(code.toString());
+
+				traverse(ast, {
+					enter(path) {
+						if (path.node.name === 'define') {
```
wincent, line 37: You could make this check a bit stricter... like:

```js
if (
    path.node.type === 'Identifier' &&
    path.node.name === 'define'
)
```

or better still:

```js
if (
    path.node.type === 'UnaryExpression' &&
    path.node.operator === 'typeof' &&
    path.node.argument.type === 'Identifier' &&
    path.node.argument.name === 'define'
)
```

The idea of using an AST-based transform is to make it precise, like a scalpel, so that you don't edit something that you don't mean to edit. So make it as precise as you can without it being brittle in the opposite sense.

To increase our confidence in this change, it would be good to grab a before and after diff just to "prove" it's not editing anything we don't want it to. Even as a once-off manual check pasted into the PR, that's still valuable

wincent (approved):
Looks reasonable to me but I'd still like to see a before-and-after diff pasted in here somewhere to document the effects of your change.

support/copyPluginDependencies.js
julien: Thanks for the suggestion @wincent, that's what I meant by 

> if someone can give me some pointers we could probably remove the entire if
but for the moment this fixes the issue.

I'm sure there's a way to make this more robust, but this is already a good start.

Here are some screenshots of the `ckeditor/plugins/codemirror/vendors/vendors.js` **before** (on the left) and **after** (on the right)

julien:
@wincent so fast to reply :joy:

julien:
btw, I'm not sure if there's a better way to show diffs, but if there is I'm interested

wincent:
> btw, I'm not sure if there's a better way to show diffs, but if there is I'm interested

I don't know if it's better but you can run `git diff` and use use a fenced code block of type `diff` to get it highlighted.

eg.

```
    ```diff
    diff --git a/greeting.txt b/greeting.txt
    new file mode 100644
    index 000000000000..b7c75fabee93
    --- /dev/null
    +++ b/greeting.txt
    @@ -0,0 +1 @@
    +hai
    ```
```

Renders as:

```diff
diff --git a/greeting.txt b/greeting.txt
new file mode 100644
index 000000000000..b7c75fabee93
--- /dev/null
+++ b/greeting.txt
@@ -0,0 +1 @@
+hai
```

julien:
@wincent I see ... In this case it creates a huge diff, I wonder if something is wrong

wincent:
One of the things about Babel is that it doesn't preserve whitespace, so you should do the diff with `git diff -b` (ignore changes in the amount of space) or `git diff -w` (ignore all whitespace changes).

I usually look at `-b` first, and if that isn't aggressive enough, `-w` (which may be too aggressive, which is why it's worth looking at both).

julien:
@wincent thanks for the tip.

Here's a diff 

```diff
diff --git a/ckeditor/plugins/codemirror/vendors/vendors.js b/ckeditor/plugins/codemirror/vendors/vendors.js
index dd1302231..87f950aff 100644
--- a/ckeditor/plugins/codemirror/vendors/vendors.js
+++ b/ckeditor/plugins/codemirror/vendors/vendors.js
@@ -1,4 +1,4 @@
-﻿(function(Q,da){"object"===typeof exports&&"undefined"!==typeof module?module.exports=da():"function"===typeof define&&define.amd?define(da):(Q=Q||self,Q.CodeMirror=da())})(this,function(){function Q(a){return new RegExp("(^|\\s)"+a+"(?:$|\\s)\\s*")}function da(a){for(var b=a.childNodes.length;0<b;--b)a.removeChild(a.firstChild);return a}function Z(a,b){return da(a).appendChild(b)}function r(a,b,c,d){a=document.createElement(a);c&&(a.className=c);d&&(a.style.cssText=d);if("string"==typeof b)a.appendChild(document.createTextNode(b));
+﻿(function(Q,da){"object"===typeof exports&&"undefined"!==typeof module?module.exports=da():"function"===__define_disabled__&&define.amd?define(da):(Q=Q||self,Q.CodeMirror=da())})(this,function(){function Q(a){return new RegExp("(^|\\s)"+a+"(?:$|\\s)\\s*")}function da(a){for(var b=a.childNodes.length;0<b;--b)a.removeChild(a.firstChild);return a}function Z(a,b){return da(a).appendChild(b)}function r(a,b,c,d){a=document.createElement(a);c&&(a.className=c);d&&(a.style.cssText=d);if("string"==typeof b)a.appendChild(document.createTextNode(b));
 else if(b)for(c=0;c<b.length;++c)a.appendChild(b[c]);return a}function y(a,b,c,d){a=r(a,b,c,d);a.setAttribute("role","presentation");return a}function A(a,b){3==b.nodeType&&(b=b.parentNode);if(a.contains)return a.contains(b);do if(11==b.nodeType&&(b=b.host),b==a)return!0;while(b=b.parentNode)}function m(){var a;try{a=document.activeElement}catch(b){a=document.body||null}for(;a&&a.shadowRoot&&a.shadowRoot.activeElement;)a=a.shadowRoot.activeElement;return a}function B(a,b){var c=a.className;Q(b).test(c)||
 (a.className+=(c?" ":"")+b)}function n(a,b){for(var c=a.split(" "),d=0;d<c.length;d++)c[d]&&!Q(c[d]).test(b)&&(b+=" "+c[d]);return b}function f(a){var b=Array.prototype.slice.call(arguments,1);return function(){return a.apply(null,b)}}function k(a,b,c){b||(b={});for(var d in a)!a.hasOwnProperty(d)||!1===c&&b.hasOwnProperty(d)||(b[d]=a[d]);return b}function I(a,b,c,d,e){null==b&&(b=a.search(/[^\s\u00a0]/),-1==b&&(b=a.length));d=d||0;for(e=e||0;;){var g=a.indexOf("\t",d);if(0>g||g>=b)return e+(b-d);
 e+=g-d;e+=c-e%c;d=g+1}}function v(a,b){for(var c=0;c<a.length;++c)if(a[c]==b)return c;return-1}function J(a,b,c){for(var d=0,e=0;;){var g=a.indexOf("\t",d);-1==g&&(g=a.length);var h=g-d;if(g==a.length||e+h>=b)return d+Math.min(h,b-e);e+=g-d;e+=c-e%c;d=g+1;if(e>=b)return d}}function G(a){for(;nc.length<=a;)nc.push(z(nc)+" ");return nc[a]}function z(a){return a[a.length-1]}function Fa(a,b){for(var c=[],d=0;d<a.length;d++)c[d]=b(a[d],d);return c}function oc(a,b,c){for(var d=0,e=c(b);d<a.length&&c(a[d])<=
@@ -326,7 +326,7 @@ T.defineMIME("text/plain","null");T.defineExtension=function(a,b){T.prototype[a]
 a.form;e=g.submit;try{var h=g.submit=function(){c();g.submit=e;g.submit();g.submit=h}}catch(l){}}b.finishInit=function(d){d.save=c;d.getTextArea=function(){return a};d.toTextArea=function(){d.toTextArea=isNaN;c();a.parentNode.removeChild(d.getWrapperElement());a.style.display="";a.form&&(R(a.form,"submit",c),b.leaveSubmitMethodAlone||"function"!=typeof a.form.submit||(a.form.submit=e))}};a.style.display="none";var f=T(function(b){return a.parentNode.insertBefore(b,a.nextSibling)},b);return f};(function(a){a.off=
 R;a.on=E;a.wheelEventPixels=Eg;a.Doc=la;a.splitLines=Wd;a.countColumn=I;a.findColumn=J;a.isWordChar=C;a.Pass=Qc;a.signal=X;a.Line=rb;a.changeEnd=Oa;a.scrollbarModel=Ve;a.Pos=w;a.cmpPos=K;a.modes=Xc;a.mimeModes=ib;a.resolveMode=qc;a.getMode=Yc;a.modeExtensions=jb;a.extendMode=jg;a.copyState=Ta;a.startState=ae;a.innerMode=Zc;a.commands=ac;a.keyMap=$b;a.keyName=Gf;a.isModifierKey=Df;a.lookupKey=xb;a.normalizeKeyMap=Qg;a.StringStream=Y;a.SharedTextMarker=Zb;a.TextMarker=Qa;a.LineWidget=Yb;a.e_preventDefault=
 ja;a.e_stopPropagation=Zd;a.e_stop=Db;a.addClass=B;a.contains=A;a.rmClass=ab;a.keyNames=Ra})(T);T.version="5.57.0";return T});
-(function(Q){"object"==typeof exports&&"object"==typeof module?Q(require("../../lib/codemirror")):"function"==typeof define&&define.amd?define(["../../lib/codemirror"],Q):Q(CodeMirror)})(function(Q){var da={autoSelfClosers:{area:!0,base:!0,br:!0,col:!0,command:!0,embed:!0,frame:!0,hr:!0,img:!0,input:!0,keygen:!0,link:!0,meta:!0,param:!0,source:!0,track:!0,wbr:!0,menuitem:!0},implicitlyClosed:{dd:!0,li:!0,optgroup:!0,option:!0,p:!0,rp:!0,rt:!0,tbody:!0,td:!0,tfoot:!0,th:!0,tr:!0},contextGrabbers:{dd:{dd:!0,
+(function(Q){"object"==typeof exports&&"object"==typeof module?Q(require("../../lib/codemirror")):"function"==__define_disabled__&&define.amd?define(["../../lib/codemirror"],Q):Q(CodeMirror)})(function(Q){var da={autoSelfClosers:{area:!0,base:!0,br:!0,col:!0,command:!0,embed:!0,frame:!0,hr:!0,img:!0,input:!0,keygen:!0,link:!0,meta:!0,param:!0,source:!0,track:!0,wbr:!0,menuitem:!0},implicitlyClosed:{dd:!0,li:!0,optgroup:!0,option:!0,p:!0,rp:!0,rt:!0,tbody:!0,td:!0,tfoot:!0,th:!0,tr:!0},contextGrabbers:{dd:{dd:!0,
 dt:!0},dt:{dd:!0,dt:!0},li:{li:!0},option:{option:!0,optgroup:!0},optgroup:{optgroup:!0},p:{address:!0,article:!0,aside:!0,blockquote:!0,dir:!0,div:!0,dl:!0,fieldset:!0,footer:!0,form:!0,h1:!0,h2:!0,h3:!0,h4:!0,h5:!0,h6:!0,header:!0,hgroup:!0,hr:!0,menu:!0,nav:!0,ol:!0,p:!0,pre:!0,section:!0,table:!0,ul:!0},rp:{rp:!0,rt:!0},rt:{rp:!0,rt:!0},tbody:{tbody:!0,tfoot:!0},td:{td:!0,th:!0},tfoot:{tbody:!0},th:{td:!0,th:!0},thead:{tbody:!0,tfoot:!0},tr:{tr:!0}},doNotIndent:{pre:!0},allowUnquoted:!0,allowMissing:!0,
 caseFold:!0},Z={autoSelfClosers:{},implicitlyClosed:{},contextGrabbers:{},doNotIndent:{},allowUnquoted:!1,allowMissing:!1,allowMissingTagName:!1,caseFold:!1};Q.defineMode("xml",function(r,y){function A(k,t){function C(f){t.tokenize=f;return f(k,t)}var H=k.next();if("\x3c"==H){if(k.eat("!"))return k.eat("[")?k.match("CDATA[")?C(n("atom","]]\x3e")):null:k.match("--")?C(n("comment","--\x3e")):k.match("DOCTYPE",!0,!0)?(k.eatWhile(/[\w\._\-]/),C(f(1))):null;if(k.eat("?"))return k.eatWhile(/[\w\._\-]/),
 t.tokenize=n("meta","?\x3e"),"meta";Ka=k.eat("/")?"closeTag":"openTag";t.tokenize=m;return"tag bracket"}if("\x26"==H)return(k.eat("#")?k.eat("x")?k.eatWhile(/[a-fA-F\d]/)&&k.eat(";"):k.eatWhile(/[\d]/)&&k.eat(";"):k.eatWhile(/[\w\.\-:]/)&&k.eat(";"))?"atom":"error";k.eatWhile(/[^&<]/);return null}function m(f,k){var t=f.next();if("\x3e"==t||"/"==t&&f.eat("\x3e"))return k.tokenize=A,Ka="\x3e"==t?"endTag":"selfcloseTag","tag bracket";if("\x3d"==t)return Ka="equals",null;if("\x3c"==t)return k.tokenize=
@@ -421,5 +421,5 @@ n=this._create_token(k.TAG_OPEN,n));return n};A.prototype._read_open_handlebars=
 f||"/"===f&&"\x3e"===this._input.peek(1))?(n=this._input.next(),"/"===f&&(n+=this._input.next()),n=this._create_token(k.TAG_CLOSE,n)):"{"===m.text[0]&&"}"===f&&"}"===this._input.peek(1)&&(this._input.next(),this._input.next(),n=this._create_token(k.TAG_CLOSE,"}}")));return n};A.prototype._read_attribute=function(f,m,n){var r=null,y="";n&&"\x3c"===n.text[0]&&("\x3d"===f?r=this._create_token(k.EQUALS,this._input.next()):'"'===f||"'"===f?(m=this._input.next(),m='"'===f?m+this.__patterns.double_quote.read():
 m+this.__patterns.single_quote.read(),r=this._create_token(k.VALUE,m)):(y=this.__patterns.attribute.read())&&(r=m.type===k.EQUALS?this._create_token(k.VALUE,y):this._create_token(k.ATTRIBUTE,y)));return r};A.prototype._is_content_unformatted=function(f){return-1===this._options.void_elements.indexOf(f)&&(-1!==this._options.content_unformatted.indexOf(f)||-1!==this._options.unformatted.indexOf(f))};A.prototype._read_raw_content=function(f,m,n){var r="";if(n&&"{"===n.text[0])r=this.__patterns.handlebars_raw_close.read();
 else if(m.type===k.TAG_CLOSE&&"\x3c"===m.opened.text[0]&&"/"!==m.text[0])if(m=m.opened.text.substr(1).toLowerCase(),"script"===m||"style"===m){if(f=this._read_comment_or_cdata(f))return f.type=k.TEXT,f;r=this._input.readUntil(new RegExp("\x3c/"+m+"[\\n\\r\\t ]*?\x3e","ig"))}else this._is_content_unformatted(m)&&(r=this._input.readUntil(new RegExp("\x3c/"+m+"[\\n\\r\\t ]*?\x3e","ig")));return r?this._create_token(k.TEXT,r):null};A.prototype._read_content_word=function(f){var m="";this._options.unformatted_content_delimiter&&
-f===this._options.unformatted_content_delimiter[0]&&(m=this.__patterns.unformatted_content_delimiter.read());m||(m=this.__patterns.word.read());if(m)return this._create_token(k.TEXT,m)};r.exports.Tokenizer=A;r.exports.TOKEN=k}]);if("function"===typeof define&&define.amd)define(["require","./beautify","./beautify-css"],function(r){var y=r("./beautify"),A=r("./beautify-css");return{html_beautify:function(m,r){return Q(m,r,y.js_beautify,A.css_beautify)}}});else if("undefined"!==typeof exports){var da=
+f===this._options.unformatted_content_delimiter[0]&&(m=this.__patterns.unformatted_content_delimiter.read());m||(m=this.__patterns.word.read());if(m)return this._create_token(k.TEXT,m)};r.exports.Tokenizer=A;r.exports.TOKEN=k}]);if("function"===__define_disabled__&&define.amd)define(["require","./beautify","./beautify-css"],function(r){var y=r("./beautify"),A=r("./beautify-css");return{html_beautify:function(m,r){return Q(m,r,y.js_beautify,A.css_beautify)}}});else if("undefined"!==typeof exports){var da=
 require("./beautify.js"),Z=require("./beautify-css.js");exports.html_beautify=function(r,y){return Q(r,y,da.js_beautify,Z.css_beautify)}}else"undefined"!==typeof window?window.html_beautify=function(r,y){return Q(r,y,window.js_beautify,window.css_beautify)}:"undefined"!==typeof global&&(global.html_beautify=function(r,y){return Q(r,y,global.js_beautify,global.css_beautify)})})();
\ No newline at end of file
```

wincent:
So there's a bug in there:

`typeof define` got changed to `__define_disabled__`

You want it to be `typeof __define_disable__`.

I hereby withdraw my approval! 😉

julien:
Yes, that probably because I was using  `path.replaceWithSourceString('__define_disabled__');` which I now updated.


Here's an updated diff 

```diff
diff --git a/ckeditor/plugins/codemirror/vendors/vendors.js b/ckeditor/plugins/codemirror/vendors/vendors.js
index dd1302231..205f273f8 100644
--- a/ckeditor/plugins/codemirror/vendors/vendors.js
+++ b/ckeditor/plugins/codemirror/vendors/vendors.js
@@ -1,4 +1,4 @@
-﻿(function(Q,da){"object"===typeof exports&&"undefined"!==typeof module?module.exports=da():"function"===typeof define&&define.amd?define(da):(Q=Q||self,Q.CodeMirror=da())})(this,function(){function Q(a){return new RegExp("(^|\\s)"+a+"(?:$|\\s)\\s*")}function da(a){for(var b=a.childNodes.length;0<b;--b)a.removeChild(a.firstChild);return a}function Z(a,b){return da(a).appendChild(b)}function r(a,b,c,d){a=document.createElement(a);c&&(a.className=c);d&&(a.style.cssText=d);if("string"==typeof b)a.appendChild(document.createTextNode(b));
+﻿(function(Q,da){"object"===typeof exports&&"undefined"!==typeof module?module.exports=da():"function"===typeof __define_disabled__&&define.amd?define(da):(Q=Q||self,Q.CodeMirror=da())})(this,function(){function Q(a){return new RegExp("(^|\\s)"+a+"(?:$|\\s)\\s*")}function da(a){for(var b=a.childNodes.length;0<b;--b)a.removeChild(a.firstChild);return a}function Z(a,b){return da(a).appendChild(b)}function r(a,b,c,d){a=document.createElement(a);c&&(a.className=c);d&&(a.style.cssText=d);if("string"==typeof b)a.appendChild(document.createTextNode(b));
 else if(b)for(c=0;c<b.length;++c)a.appendChild(b[c]);return a}function y(a,b,c,d){a=r(a,b,c,d);a.setAttribute("role","presentation");return a}function A(a,b){3==b.nodeType&&(b=b.parentNode);if(a.contains)return a.contains(b);do if(11==b.nodeType&&(b=b.host),b==a)return!0;while(b=b.parentNode)}function m(){var a;try{a=document.activeElement}catch(b){a=document.body||null}for(;a&&a.shadowRoot&&a.shadowRoot.activeElement;)a=a.shadowRoot.activeElement;return a}function B(a,b){var c=a.className;Q(b).test(c)||
 (a.className+=(c?" ":"")+b)}function n(a,b){for(var c=a.split(" "),d=0;d<c.length;d++)c[d]&&!Q(c[d]).test(b)&&(b+=" "+c[d]);return b}function f(a){var b=Array.prototype.slice.call(arguments,1);return function(){return a.apply(null,b)}}function k(a,b,c){b||(b={});for(var d in a)!a.hasOwnProperty(d)||!1===c&&b.hasOwnProperty(d)||(b[d]=a[d]);return b}function I(a,b,c,d,e){null==b&&(b=a.search(/[^\s\u00a0]/),-1==b&&(b=a.length));d=d||0;for(e=e||0;;){var g=a.indexOf("\t",d);if(0>g||g>=b)return e+(b-d);
 e+=g-d;e+=c-e%c;d=g+1}}function v(a,b){for(var c=0;c<a.length;++c)if(a[c]==b)return c;return-1}function J(a,b,c){for(var d=0,e=0;;){var g=a.indexOf("\t",d);-1==g&&(g=a.length);var h=g-d;if(g==a.length||e+h>=b)return d+Math.min(h,b-e);e+=g-d;e+=c-e%c;d=g+1;if(e>=b)return d}}function G(a){for(;nc.length<=a;)nc.push(z(nc)+" ");return nc[a]}function z(a){return a[a.length-1]}function Fa(a,b){for(var c=[],d=0;d<a.length;d++)c[d]=b(a[d],d);return c}function oc(a,b,c){for(var d=0,e=c(b);d<a.length&&c(a[d])<=
@@ -326,7 +326,7 @@ T.defineMIME("text/plain","null");T.defineExtension=function(a,b){T.prototype[a]
 a.form;e=g.submit;try{var h=g.submit=function(){c();g.submit=e;g.submit();g.submit=h}}catch(l){}}b.finishInit=function(d){d.save=c;d.getTextArea=function(){return a};d.toTextArea=function(){d.toTextArea=isNaN;c();a.parentNode.removeChild(d.getWrapperElement());a.style.display="";a.form&&(R(a.form,"submit",c),b.leaveSubmitMethodAlone||"function"!=typeof a.form.submit||(a.form.submit=e))}};a.style.display="none";var f=T(function(b){return a.parentNode.insertBefore(b,a.nextSibling)},b);return f};(function(a){a.off=
 R;a.on=E;a.wheelEventPixels=Eg;a.Doc=la;a.splitLines=Wd;a.countColumn=I;a.findColumn=J;a.isWordChar=C;a.Pass=Qc;a.signal=X;a.Line=rb;a.changeEnd=Oa;a.scrollbarModel=Ve;a.Pos=w;a.cmpPos=K;a.modes=Xc;a.mimeModes=ib;a.resolveMode=qc;a.getMode=Yc;a.modeExtensions=jb;a.extendMode=jg;a.copyState=Ta;a.startState=ae;a.innerMode=Zc;a.commands=ac;a.keyMap=$b;a.keyName=Gf;a.isModifierKey=Df;a.lookupKey=xb;a.normalizeKeyMap=Qg;a.StringStream=Y;a.SharedTextMarker=Zb;a.TextMarker=Qa;a.LineWidget=Yb;a.e_preventDefault=
 ja;a.e_stopPropagation=Zd;a.e_stop=Db;a.addClass=B;a.contains=A;a.rmClass=ab;a.keyNames=Ra})(T);T.version="5.57.0";return T});
-(function(Q){"object"==typeof exports&&"object"==typeof module?Q(require("../../lib/codemirror")):"function"==typeof define&&define.amd?define(["../../lib/codemirror"],Q):Q(CodeMirror)})(function(Q){var da={autoSelfClosers:{area:!0,base:!0,br:!0,col:!0,command:!0,embed:!0,frame:!0,hr:!0,img:!0,input:!0,keygen:!0,link:!0,meta:!0,param:!0,source:!0,track:!0,wbr:!0,menuitem:!0},implicitlyClosed:{dd:!0,li:!0,optgroup:!0,option:!0,p:!0,rp:!0,rt:!0,tbody:!0,td:!0,tfoot:!0,th:!0,tr:!0},contextGrabbers:{dd:{dd:!0,
+(function(Q){"object"==typeof exports&&"object"==typeof module?Q(require("../../lib/codemirror")):"function"==typeof __define_disabled__&&define.amd?define(["../../lib/codemirror"],Q):Q(CodeMirror)})(function(Q){var da={autoSelfClosers:{area:!0,base:!0,br:!0,col:!0,command:!0,embed:!0,frame:!0,hr:!0,img:!0,input:!0,keygen:!0,link:!0,meta:!0,param:!0,source:!0,track:!0,wbr:!0,menuitem:!0},implicitlyClosed:{dd:!0,li:!0,optgroup:!0,option:!0,p:!0,rp:!0,rt:!0,tbody:!0,td:!0,tfoot:!0,th:!0,tr:!0},contextGrabbers:{dd:{dd:!0,
 dt:!0},dt:{dd:!0,dt:!0},li:{li:!0},option:{option:!0,optgroup:!0},optgroup:{optgroup:!0},p:{address:!0,article:!0,aside:!0,blockquote:!0,dir:!0,div:!0,dl:!0,fieldset:!0,footer:!0,form:!0,h1:!0,h2:!0,h3:!0,h4:!0,h5:!0,h6:!0,header:!0,hgroup:!0,hr:!0,menu:!0,nav:!0,ol:!0,p:!0,pre:!0,section:!0,table:!0,ul:!0},rp:{rp:!0,rt:!0},rt:{rp:!0,rt:!0},tbody:{tbody:!0,tfoot:!0},td:{td:!0,th:!0},tfoot:{tbody:!0},th:{td:!0,th:!0},thead:{tbody:!0,tfoot:!0},tr:{tr:!0}},doNotIndent:{pre:!0},allowUnquoted:!0,allowMissing:!0,
 caseFold:!0},Z={autoSelfClosers:{},implicitlyClosed:{},contextGrabbers:{},doNotIndent:{},allowUnquoted:!1,allowMissing:!1,allowMissingTagName:!1,caseFold:!1};Q.defineMode("xml",function(r,y){function A(k,t){function C(f){t.tokenize=f;return f(k,t)}var H=k.next();if("\x3c"==H){if(k.eat("!"))return k.eat("[")?k.match("CDATA[")?C(n("atom","]]\x3e")):null:k.match("--")?C(n("comment","--\x3e")):k.match("DOCTYPE",!0,!0)?(k.eatWhile(/[\w\._\-]/),C(f(1))):null;if(k.eat("?"))return k.eatWhile(/[\w\._\-]/),
 t.tokenize=n("meta","?\x3e"),"meta";Ka=k.eat("/")?"closeTag":"openTag";t.tokenize=m;return"tag bracket"}if("\x26"==H)return(k.eat("#")?k.eat("x")?k.eatWhile(/[a-fA-F\d]/)&&k.eat(";"):k.eatWhile(/[\d]/)&&k.eat(";"):k.eatWhile(/[\w\.\-:]/)&&k.eat(";"))?"atom":"error";k.eatWhile(/[^&<]/);return null}function m(f,k){var t=f.next();if("\x3e"==t||"/"==t&&f.eat("\x3e"))return k.tokenize=A,Ka="\x3e"==t?"endTag":"selfcloseTag","tag bracket";if("\x3d"==t)return Ka="equals",null;if("\x3c"==t)return k.tokenize=
@@ -421,5 +421,5 @@ n=this._create_token(k.TAG_OPEN,n));return n};A.prototype._read_open_handlebars=
 f||"/"===f&&"\x3e"===this._input.peek(1))?(n=this._input.next(),"/"===f&&(n+=this._input.next()),n=this._create_token(k.TAG_CLOSE,n)):"{"===m.text[0]&&"}"===f&&"}"===this._input.peek(1)&&(this._input.next(),this._input.next(),n=this._create_token(k.TAG_CLOSE,"}}")));return n};A.prototype._read_attribute=function(f,m,n){var r=null,y="";n&&"\x3c"===n.text[0]&&("\x3d"===f?r=this._create_token(k.EQUALS,this._input.next()):'"'===f||"'"===f?(m=this._input.next(),m='"'===f?m+this.__patterns.double_quote.read():
 m+this.__patterns.single_quote.read(),r=this._create_token(k.VALUE,m)):(y=this.__patterns.attribute.read())&&(r=m.type===k.EQUALS?this._create_token(k.VALUE,y):this._create_token(k.ATTRIBUTE,y)));return r};A.prototype._is_content_unformatted=function(f){return-1===this._options.void_elements.indexOf(f)&&(-1!==this._options.content_unformatted.indexOf(f)||-1!==this._options.unformatted.indexOf(f))};A.prototype._read_raw_content=function(f,m,n){var r="";if(n&&"{"===n.text[0])r=this.__patterns.handlebars_raw_close.read();
 else if(m.type===k.TAG_CLOSE&&"\x3c"===m.opened.text[0]&&"/"!==m.text[0])if(m=m.opened.text.substr(1).toLowerCase(),"script"===m||"style"===m){if(f=this._read_comment_or_cdata(f))return f.type=k.TEXT,f;r=this._input.readUntil(new RegExp("\x3c/"+m+"[\\n\\r\\t ]*?\x3e","ig"))}else this._is_content_unformatted(m)&&(r=this._input.readUntil(new RegExp("\x3c/"+m+"[\\n\\r\\t ]*?\x3e","ig")));return r?this._create_token(k.TEXT,r):null};A.prototype._read_content_word=function(f){var m="";this._options.unformatted_content_delimiter&&
-f===this._options.unformatted_content_delimiter[0]&&(m=this.__patterns.unformatted_content_delimiter.read());m||(m=this.__patterns.word.read());if(m)return this._create_token(k.TEXT,m)};r.exports.Tokenizer=A;r.exports.TOKEN=k}]);if("function"===typeof define&&define.amd)define(["require","./beautify","./beautify-css"],function(r){var y=r("./beautify"),A=r("./beautify-css");return{html_beautify:function(m,r){return Q(m,r,y.js_beautify,A.css_beautify)}}});else if("undefined"!==typeof exports){var da=
+f===this._options.unformatted_content_delimiter[0]&&(m=this.__patterns.unformatted_content_delimiter.read());m||(m=this.__patterns.word.read());if(m)return this._create_token(k.TEXT,m)};r.exports.Tokenizer=A;r.exports.TOKEN=k}]);if("function"===typeof __define_disabled__&&define.amd)define(["require","./beautify","./beautify-css"],function(r){var y=r("./beautify"),A=r("./beautify-css");return{html_beautify:function(m,r){return Q(m,r,y.js_beautify,A.css_beautify)}}});else if("undefined"!==typeof exports){var da=
 require("./beautify.js"),Z=require("./beautify-css.js");exports.html_beautify=function(r,y){return Q(r,y,da.js_beautify,Z.css_beautify)}}else"undefined"!==typeof window?window.html_beautify=function(r,y){return Q(r,y,window.js_beautify,window.css_beautify)}:"undefined"!==typeof global&&(global.html_beautify=function(r,y){return Q(r,y,global.js_beautify,global.css_beautify)})})();
\ No newline at end of file
```

wincent:
lg2m

carloslancha (approved):
lgtm

PR #142, opened 2020-11-26
fix: babel deps are devdeps

carloslancha:
Didn't catch in https://github.com/liferay/liferay-ckeditor/pull/141


julien (approved):
Thanks

wincent:
> babel deps are devdeps

oops

PR #146, opened 2020-12-18
fix: LPS-124728 Avoid using unsupported APIs in IE11

julien:
We were using `URL`, `URLSearchParams` and `Object.entries`, 
which aren't supported in IE11 . In theory we have polyfills for these 
but apparently when `getUrl` is executed these polyfills aren't loaded yet.

So this patch reverts a previous patch and adds more .

Tested agaisn't [LPS-124728](https://issues.liferay.com/browse/LPS-124728) and [LPS-112982](https://issues.liferay.com/browse/LPS-112982)

### DIFF
--- a/patches/0007-LPS-124728-Avoid-breaking-IE11.patch
+++ b/patches/0007-LPS-124728-Avoid-breaking-IE11.patch
@@ -1,0 +1,61 @@
+From d0bf70988e63c2c84076d7418eb6cfe0edc438de Mon Sep 17 00:00:00 2001
+From: Julien Castelain <julien.castelain@liferay.com>
+Date: Mon, 21 Dec 2020 09:12:53 +0100
+Subject: [PATCH] LPS-124728 Avoid breaking IE11
+
+IE11 Doesn't support URL, URLSearchParams and Object.entries
+In theory though we have a pollyfill for these but it seems that
+they aren't loaded when getUrl is executed.
+---
+ core/ckeditor_base.js | 30 ++++++++++++++++++------------
+ 1 file changed, 18 insertions(+), 12 deletions(-)
+
+diff --git a/core/ckeditor_base.js b/core/ckeditor_base.js
+index 6d9fcbc03..59b3e4a0b 100644
+--- a/core/ckeditor_base.js
++++ b/core/ckeditor_base.js
+@@ -171,26 +171,32 @@ if ( !window.CKEDITOR ) {
+ 			 * @returns {String} The full URL.
+ 			 */
+ 			getUrl: function( resource ) {
+-				var resourceUrl = resource.startsWith(this.basePath)
+-					? new URL(resource)
+-					: new URL(this.basePath + resource);
++				function set(key, value) {
++					if (!resource.match(new RegExp('[&?]' + key + '\\b'))) {
++						resource += (resource.indexOf('?') >= 0 ? '&' : '?') + key + '=' + value;
++					}
++				}
+ 
+-				if (!resourceUrl.toString().endsWith('/')) {
+-					if (this.timestamp && !resourceUrl.searchParams.has('t')) {
+-						resourceUrl.searchParams.set('t', this.timestamp);
++				// If this is not a full or absolute path.
++				if (resource.indexOf(':/') === -1 && resource.indexOf('/') !== 0) {
++					resource = this.basePath + resource;
++				}
++
++				if (resource.charAt(resource.length - 1) !== '/') {
++					if (this.timestamp) {
++						set('t', this.timestamp);
+ 					}
+ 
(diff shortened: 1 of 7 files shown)


patches/0007-LPS-124728-Avoid-breaking-IE11.patch
```diff
@@ -0,0 +1,63 @@
...
++++ b/core/ckeditor_base.js
+@@ -171,26 +171,31 @@ if ( !window.CKEDITOR ) {
+ 			 * @returns {String} The full URL.
+ 			 */
+ 			getUrl: function( resource ) {
+-				var resourceUrl = resource.startsWith(this.basePath)
+-					? new URL(resource)
+-					: new URL(this.basePath + resource);
++				// If this is not a full or absolute path.
++				if ( resource.indexOf( ':/' ) == -1 && resource.indexOf( '/' ) !== 0 )
```
wincent, line 25: Excess whitespace here inside the parens, inconsistent with other parts of the patch (I know this is partly resurrecting old code from prior to a previous patch, but we should still be consistent). Likewise, using `==` and `!==` inconsistently; we should in [general favor the "triple"/"strict" variants](https://github.com/liferay/liferay-frontend-projects/issues/14).

(non-blocking feedback)

patches/0007-LPS-124728-Avoid-breaking-IE11.patch
```diff
@@ -0,0 +1,63 @@
...
+-					if (this.timestamp && !resourceUrl.searchParams.has('t')) {
+-						resourceUrl.searchParams.set('t', this.timestamp);
++				// Add the timestamp, except for directories.
++				if ( this.timestamp && resource.charAt( resource.length - 1 ) != '/' && !( /[&?]t=/ ).test( resource ) )
++					resource += ( resource.indexOf( '?' ) >= 0 ? '&' : '?' ) + 't=' + this.timestamp;
++
++				Object.keys(CKEDITOR.ADDITIONAL_RESOURCE_PARAMS).forEach(function(key) {
++					var value = CKEDITOR.ADDITIONAL_RESOURCE_PARAMS[key];
++
++					if ((resource.charAt(resource.length - 1) === '/') || (/\.css$/).test(resource) && key === 'languageId') {
```
wincent, line 38: Rather than repeating this `charAt()` check for every param, maybe you should do it once, like the old version did.

ie.

```js
if (resource.charAt(resource.length - 1) !== '/') {
    if (this.timestamp && !/[&?]t=/.test(resource)) {
        resource += (resource.indexOf('?') >= 0 ? '&' : '?') + 't=' + this.timestamp;
    }

    Object.keys(CKEDITOR.ADDITIONAL_RESOURCE_PARAMS).forEach(function(key) {
        var value = CKEDITOR.ADDITIONAL_RESOURCE_PARAMS[key];

        if (/\.css$/.test(resource) && key === 'languageId') {
            return;
        }

        var regexp = new RegExp('[&?]' + key);

        if (!regexp.test(resource)) {
            resource += (resource.indexOf('?') >= 0 ? '&' : '?') + key + '=' + value;
        }
    });
}

return resource;
```

It is also tempting to factor out the repeated ternaries... and fix a bug there with that `new RegExp` call which given key "foo" will create a regexp that matches `?foo`, `?foobar`, and `?foobarbaz` etc...

```js
function set(key, value) {
    if (!resource.match(new RegExp('[&?]' + key + '\\b'))) {
        resource += (resource.indexOf('?') >= 0 ? '&' : '?') + key + '=' + value;
    }
}

if (resource.charAt(resource.length - 1) !== '/') {
    if (this.timestamp) {
        set('t', this.timestamp);
    }

    Object.keys(CKEDITOR.ADDITIONAL_RESOURCE_PARAMS).forEach(function(key) {
        var value = CKEDITOR.ADDITIONAL_RESOURCE_PARAMS[key];

        if (!/\.css$/.test(resource) || !key === 'languageId') {
            set(key, value);
        }
    });
}

return resource;
```

wincent (commented):
A suggestion to think about inline. Non-blocking, apart from the `new RegExp()` construction which I think probably qualifies as a bug.

patches/0007-LPS-124728-Avoid-breaking-IE11.patch
julien: Thanks. Updated in [a11ce92](https://github.com/liferay/liferay-ckeditor/pull/146/commits/a11ce921e841208128bb7f6511b97455bc9bf2c7)

jbalsas:
Hey @julien, feel free to merge and release if this is blocking DXP releases. I assume you've verified this works as expected, so if you're confident, let's start getting this through the process!

julien:
Yes, I've tested it. Going to prepare the release now.

PR #148, opened 2021-01-04
fix: LPS-125441 CKEditor font size selector flickers on mouse over

NemethNorbert:
Hey,

[LPS-125441](https://issues.liferay.com/browse/LPS-125441)
Our ckeditor skin had a flickering, as the focus and active state of the elements did not contain the necessary padding and margin values. This issue is not reproducible on master due to our toolbar change, however on 7.2.x where the toolbar still holds the same plugins, it is reproducible.

Please review my changes

### DIFF
--- a/skins/moono-lexicon/richcombo.css
+++ b/skins/moono-lexicon/richcombo.css
@@ -196,8 +196,8 @@ a.cke_combo_button {
 .cke_toolbar_break + .cke_toolbar > .cke_toolbar_start + .cke_combo_off a.cke_combo_button:focus,
 .cke_toolbar_break + .cke_toolbar > .cke_toolbar_start + .cke_combo_off a.cke_combo_button:active {
 	/* If first combo in a row do not use negative margin so it is aligned with other rows on hover. */
-	padding: 0;
-	margin: 0;
+	padding: 6px 18px;
+	margin: 0 4px 0 0;
 }
 
 .cke_hc


NemethNorbert:
Hey @carloslancha,

When can you move forward with the review?

Thanks,

pat270:
@NemethNorbert This LGTM. I don't know where to get or deploy hotfixes to bundles. I was able to simulate by replacing editor.css with the monolisa skin fron this repo. This is what I got:
ckechanges

diegonvs:
Hey, I took a time taking a look on it and looks like this error cannot be reproduced on current master because the current version of liferay-ckeditor there is applying the following classes:

image

However, on 7.2.x, We are applying different styles:

image

taking a look on the [7.2.x liferay-ckeditor's version](https://github.com/liferay/liferay-portal/blob/7.2.x/modules/apps/frontend-editor/frontend-editor-ckeditor-web/package.json#L3) is different from the [master](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/package.json#L5). Taking a look more closely on [4.11.4-liferay.2 version](https://github.com/liferay/liferay-ckeditor/tree/v4.11.4-liferay.2/ckeditor/skins) We don't have moono-lexicon styles applied yet and it's very weird loading moono-lexicon on this DXP version without having it.

Your fix looks okay for the 4.11.4-liferay.2 but I think it was already fixed on master but I would call @jbalsas @carloslancha for reviewing it because I may missing something

jbalsas:
@antonio-ortega, could you help assess this? 

> We don't have moono-lexicon styles applied yet and it's very weird loading moono-lexicon on this DXP version without having it.

I think this is the key, `7.2.x` should not be affected and should continue to use the same default skin that has been using. Feels like something is missing or causing the `lexicon` skin to be loaded instead which shouldn't be the case.

julien:
I checked on **liferay-portal-ee/7.2.x**, and these changes do fix the flickering issue, but as @jbalsas commented, I want to see and understand what's happening in **liferay-portal/7.2.x** so I'll compile and check that now.

@diegonvs, I think the confusion comes from the fact that [liferay/7.2.x uses liferay-ckeditor@4.11.4-liferay.2](https://github.com/liferay/liferay-portal/blob/ded0e9390637985231e962e4ad4cfa4639eabb26/modules/apps/frontend-editor/frontend-editor-ckeditor-web/package.json#L3) while [liferay-portal-ee/7.2.x uses liferay-ckeditor@4.14.1-liferay.14](https://github.com/liferay/liferay-portal-ee/blob/d352872096f3ddcb5cf652bf31ed897e24b2cbb1/modules/apps/frontend-editor/frontend-editor-ckeditor-web/package.json#L5)

julien:
This is what I see in web content with [lifery-portal CE 7.2.1 GA 2](https://sourceforge.net/projects/lportal/files/Liferay%20Portal/7.2.1%20GA2/liferay-ce-portal-tomcat-7.2.1-ga2-20191111141448326.tar.gz/download) (downloaded from sourceforge, I was unable to compile the `7.2.x` branch in my local clone) - but 7.2.x CE doesn't seem affected.



The `moono-lexicon` skin isn't loaded or used by the editor and there's no flickering problem either on the font size combobox

This is what I see in web content with [liferay-portal-ee/7.2.x](https://github.com/liferay/liferay-portal-ee/tree/7.2.x/)



The `moono-lexicon` skin is being used by the editor, and although it's not visible in the screenshot, there is a flickering issue.

antonio-ortega:
Hi guys,

Not sure if I can provide any kind of help here because I think @julien has already clarified it but just in case:

- Our Liferay DXP 7.2 is built from [liferay-portal-ee/tree/7.2.x](https://github.com/liferay/liferay-portal-ee/tree/7.2.x). CKEditor version in that branch is [4.14.1-liferay.14](https://github.com/liferay/liferay-portal-ee/blob/7.2.x/modules/apps/frontend-editor/frontend-editor-ckeditor-web/package.json#L5). Since we [updated](https://github.com/liferay/liferay-ckeditor/releases/tag/v4.13.1-liferay.6) the skin in 4.14.1-liferay.6, we are using the new skin in Liferay DXP 7.2. This change happened in [LPS-121241](https://issues.liferay.com/browse/LPS-121241) where we [upgraded](https://github.com/liferay/liferay-portal-ee/commit/193acf3454cd4609a8c0e35e3a87bff7669622d9) from 4.13.1-liferay.5 to 4.14.1-liferay.10 in order to fix an XSS issue. It's been delivered in [fixpack-dxp-9-7210](https://customer.liferay.com/en/downloads/-/download/liferay-dxp-7-2-fix-pack-9).
- Our Liferay CE 7.2 is built from [liferay-portal/tree/7.2.x](https://github.com/liferay/liferay-portal/tree/7.2.x), as you can see the version is still [4.14.1-liferay.2](https://github.com/liferay/liferay-portal/blob/7.2.x/modules/apps/frontend-editor/frontend-editor-ckeditor-web/package.json#L3), prior to the skin change.

And that's the reason why the issue is reproducible in a DXP bundle but not in a CE bundle.

Thanks.

Regards.

jbalsas:
Thanks for the analysis, @antonio-ortega!

Based on this, [Update CKEditor to v4.14.1](https://issues.liferay.com/browse/LPS-121241) wasn't backported properly. It should've added code to make sure that the skin used was not `lexicon` but rather `moono`. I would suggest adding that code in `7.2.x` to get it back to the old behaviour and avoid the visual changes in the editor there (as well as in older branches if it ever gets backported further).

That is, unless Support prefers to go with the visual change in CKEditor already 🤷‍♂️

carloslancha:
Thx guys for taking care of this
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia4.giphy.com%2Fmedia%2F26FLdmIp6wJr91JAI%2Fgiphy.gif&token=love" data-gitmeme-token="love"></a>

antonio-ortega:
Hi again,

I agree [Update CKEditor to v4.14.1](https://issues.liferay.com/browse/LPS-121241) wasn't backported properly, not only because of the skin change, mainly because of the missing plugins we are already working on bring back.

Said that, and regarding current customer's complains are about missing plugins or UX defects (like this one) and we have already delivered this new skin in official fixpack, I'd prefer to keep it, since using the same skin than master could help us in the future with bugs/backports related to skin, we are providing a more modern look and it's not being as a perceived bug by customers.

So talking about this issue and unless you disagree, I think we could move forward the solution.

Thanks.

Regards.

jbalsas:
> So talking about this issue and unless you disagree, I think we could move forward the solution.

Sounds good to me, I would just want to make sure that this does not cause issues in `master`. It worries me a bit that we might end up using a lot of time fixing skin issues, specially since in `7.2.x` the usage of AlloyEditor is still massive and that will load other skins which can cause conflicts. You will need to track down other fixes we added in `7.3.x` so when moving from a portlet with AlloyEditor to one with CKEditor (and vice-versa) won't cause weird styling problems.

NemethNorbert:
Hey guys,
Thank you for the review, if I understood the conclusion we are green to move forward with this fix.
If I'm right, could you guys move the PR forward please? @carloslancha @antonio-ortega

Thank you <3

antonio-ortega:
Hi @NemethNorbert ,

I don't have the permissions to move fixes forward in this repo.

Regards.

julien:
@antonio-ortega, @NemethNorbert  let me take care of this
I'll just re-test in master (again) to make sure nothing breaks

julien:
Just wanted to confirm that `master` isn't affected by this change, and it fixes the bug in `7.2.x` - so I'm merging this and going to prepare a release.

wincent:
> I don't have the permissions to move fixes forward in this repo.

Maybe these privileges can be arranged @antonio-ortega...

troll

PR #152, opened 2021-02-10
fix: Restore htmlwriter to have beautification in our source view

NemethNorbert:
Hey,

[LPS-126202](https://issues.liferay.com/browse/LPS-126202)
By restoring [htmlwriter](https://ckeditor.com/cke4/addon/htmlwriter), the beautification issue is resolved in AlloyEditor.

This issue is not reproducible with ckeditor at the moment as we used _js-beautify_ to beautify the html when switching to source mode in our new source editor plugin -> [LPS-118290](https://issues.liferay.com/browse/LPS-118290)

Restoring the htmlwriter plugin resolves the issue in AlloyEditor and it's a battle tested solution we used for years.

Please review my changes, thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,6 @@
 		"@babel/traverse": "^7.12.9",
 		"@clayui/css": "^3.14.0",
 		"codemirror": "^5.57.0",
-		"js-beautify": "^1.13.0",
 		"prettier": "2.0.4"
 	},
 	"scripts": {
--- a/plugins/codemirror/deps.json
+++ b/plugins/codemirror/deps.json
@@ -8,9 +8,6 @@
         "codemirror": [
             "lib/codemirror.js",
             "mode/xml/xml.js"
-        ],
-        "js-beautify": [
-            "js/lib/beautify-html.js"
         ]
     }
 }
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -44,7 +44,7 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 
 			var editor = dialog.getParentEditor();
 
-			codeMirrorEditor.setValue(html_beautify(editor.getData(true)));
+			codeMirrorEditor.setValue(editor.getData(true));
 
 			var preview = dialog
 				.getContentElement('main', 'preview')
@@ -246,7 +246,7 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 			var data = editor.getData();
 
 			if (codeMirrorEditor && codeMirrorEditor.getValue() !== data) {
-				codeMirrorEditor.setValue(html_beautify(data));
+				codeMirrorEditor.setValue(data);
 			}
 		},
 	};
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -26,7 +26,7 @@
 
 				var oldData = editor.getData(1);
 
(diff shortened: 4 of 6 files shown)


julien:
Hi @NemethNorbert, thanks for sending this contribution.

This sounds like an interesting idea. I was wondering if we couldn't get rid of js-beautify in the codemirror plugin since this is going to be enabled.

NemethNorbert:
Hey @julien,

I did not test that, but I'd believe we should be able to just do that

julien:
Let me fetch this branch and give it a try.

julien:
@NemethNorbert here's what I see. 

The difference is minimal, but still worth talking about (in my opinion)

With `html_beautify` (i.e. right now)



With `htmlwriter` (i.e. these changes)



As you'll see the "line height" is bigger with `htmlwriter`, I don't know if that's an issue but that's the only visible change I saw.

In terms of "size" changes, when running the default build (i.e. minified):

- With `html_beautify`, the size of the `ckeditor/ckeditor.js` script is 600K    
- With `htmlwriter`, the size of the `ckeditor/ckeditor.js` script is 604K    
       (I edited the files to make sure `html_beautify` isn't included)

I expected it to be the opposite, so you might want to re-check that.

In any case, we might want to leave this open for discussion, because in terms of look and feel it does change slightly, so I'll ping @drakonux to have his opinion too.

jbalsas:
If we can get rid of an external dependency and use something ootb, I think that's desirable, even if the behaviour is slightly different.

julien:
@jbalsas yeah that's true, the fact that `htmlwriter` is something that we get ootb is an advantage.

NemethNorbert:
I'd say in this case its better to ship our ckeditor with htmlwriter included, as this is the same behaviour they would see in other ckeditors as far as I understand. Having said that, I would prefer the experience provided by js-beautify as a user.

On the other hand htmlwriter is [configurable](https://ckeditor.com/docs/ckeditor4/latest/features/output_format.html), I think  we should be able to remove the extra lines between the html elements

julien:
@NemethNorbert you're right, by configuring it, we should be able to get more or less the same look and feel so I guess we're covered.

Since we all agree that this is the good direction, would you be able to disable/remove `html_beautify` from the `codemirror` plugin or do you want us to take care of that?

I thought that this could be done as part of this PR, but it doesn't have to.

NemethNorbert:
Hey @julien,

I can remove it, but it would be better if we also provide the configuration for htmlwriter when we do that. Should we do all that on this PR?

julien:
@NemethNorbert yes that sounds even better, if you think it's possible.

NemethNorbert:
> @NemethNorbert yes that sounds even better, if you think it's possible.

@julien sure, I have to wrap up some other tasks today, but after that I will get right back to this one and see what I can do.

Cheers

julien:
@NemethNorbert no problem, and if you need help don't hesitate - thanks!

NemethNorbert:
Hey @julien,

I updated the PR with the removal of js-beautify. I also went ahead and tried to configure the htmlwriter to remove those extra line breaks, but could not make it work yet. I'd like to suggest to move ahead with this PR as is, and work on the htmlwriter configuration on a separate ticket. The resolution of the AlloyEditor source issue is getting more and more urgent on the support side and we can't wait more.  Thank you

julien:
Hi @NemethNorbert 

Thanks. Yes that sounds reasonable. Let me just re-fetch and test to double check and I'll merge if it's OK.

julien:
LGTM

PR #154, opened 2021-03-04
fix: update codemirror plugin

### LINKED ISSUE #153: It's not possible to track changes in source mode due to Codemirror plugin
**Steps to reproduce**

1. Start Liferay (master branch).
2. Create a simple mvc-portlet which make use of liferay-ui:input-editor taglib and specify some function to onChangeMethod. I mean, something like this in your *view.jsp*: 
```
<%@ include file="/init.jsp" %>

<p>
	<b><liferay-ui:message key="mymvc.caption"/></b>

	<h1>PTR-2272</h1>
	<liferay-ui:input-editor 
		editorName="ckeditor" 
		name="PTR2272Editor"
		onChangeMethod="onChangeEditor"
	/>
</p>

<aui:script>
	function <portlet:namespace />onChangeEditor(html) {
		console.log("On change editor has been called.");
	}
</aui:script>
```
You can use the project attached in file [my-mvc-portlet-project_LPS127948](https://issues.liferay.com/secure/attachment/395813/my-mvc-portlet-project_LPS127948.zip), which already includes the .jar module inside _build/lib_ folder.
3. Deploy the generated jar module on the Portal.
4. Deploy this new portlet in your home page
5. Open browser console.
6. Start typing in the editor. Assert there is a message "On change editor has been called" every time you type something (considering debounce).
7. Move to "Source" view and keep typing.

**Expected behavior**
You keep seeing the same kind of messages.

**Actual behavior**
You stop seeing messages in browser console.

**Additional Information**
This is happening since v4.14.1-liferay.9 where we included CodeMirror plugin to edit text in source view.

julien:
Add an editable to the codemirror plugin, so that it also dispatches
the "change" event in source mode

This was copied directly from the original `sourcearea` plugin,
from [here](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/plugins/sourcearea/plugin.js#L48) and [here](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/plugins/sourcearea/plugin.js#L103-L133)

**Preview**

Fixes #153

### DIFF
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -13,7 +13,10 @@
 				var textarea = contentsSpace
 					.getDocument()
 					.createElement('textarea');
+
 				contentsSpace.append(textarea);
+
+				editor.editable(new codeMirrorEditable(editor, textarea));
 
 				instance.codeMirrorEditor = CodeMirror.fromTextArea(
 					textarea.$,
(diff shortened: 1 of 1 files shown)


julien:
cc @antonio-ortega

plugins/codemirror/plugin.js
```diff
@@ -13,8 +13,13 @@
 				var textarea = contentsSpace
 					.getDocument()
 					.createElement('textarea');
+
 				contentsSpace.append(textarea);
 
+				var editable = editor.editable(
```
wincent, line 19: I don't think you should assign this to `editable`, seeing as you never reference it anywhere else.

wincent (approved):
`refactor`? seems like a `fix` to me

I don't really know how this works, but your gif shows it working, so...

crossed-fingers

plugins/codemirror/plugin.js
julien: I blindly copied an pastad from CKEditor's code without really thinking about it, but you're right it's not used so not needed, will update that.

And I might as well change the commit message to say `fix`

julien: Updated in [https://github.com/liferay/liferay-ckeditor/commit/cd360019946b67e733d2046e6813df44040ef3af](https://github.com/liferay/liferay-ckeditor/commit/cd360019946b67e733d2046e6813df44040ef3af), I also changed the "prefix" in of the commit message

PR #155, opened 2021-03-10
fix: remove css classes as well when changing style

NemethNorbert:
Hey,
[LPS-127732](https://issues.liferay.com/browse/LPS-127732) (we can use this to update the ckeditor version in Liferay),

This issue was introduced when [this issue](https://github.com/ckeditor/ckeditor4/issues/4008) was fixed in ckeditor.
by this commit: https://github.com/ckeditor/ckeditor4/commit/aa007008415bc6c826d53ff5f6861d4f400da736#diff-9bccca5f3796dd5ea08364b02b6a6fc2bf7a6c6fb1e908c6be8f538e3f503e4b

I only changed
`range.enlarge( CKEDITOR.ENLARGE_INLINE );`
to
`range.enlarge( CKEDITOR.ENLARGE_ELEMENT );`

This resolves our current issue described in the LPS while it does not reintroduce the issue reported with ckeditor.

There was a PTR opened as well, and I followed up on the findings made there: https://issues.liferay.com/browse/PTR-2270

Please review it,

Thanks

/cc @antonio-ortega

### DIFF
--- a/plugins/removeformat/plugin.js
+++ b/plugins/removeformat/plugin.js
@@ -1,0 +1,192 @@
+/**
+ * @license Copyright (c) 2003-2020, CKSource - Frederico Knabben. All rights reserved.
+ * For licensing, see LICENSE.md or https://ckeditor.com/legal/ckeditor-oss-license
+ */
+
+CKEDITOR.plugins.add( 'removeformat', {
+	// jscs:disable maximumLineLength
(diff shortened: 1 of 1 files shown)


julien:
@NemethNorbert, I first need to check the ticket to understand what's the intent here, but I had two doubts: 

1 - Why aren't we patching the original [`removeformat`](https://github.com/ckeditor/ckeditor4/tree/eb2d11644a796cd13cd047cd55541430359f9317/plugins/removeformat) plugin?

2 - ci (github actions) failed because of the source formatting, just so you are aware.

Thanks

Issue #157, opened 2021-03-12
Evaluate risks of the way we build using plugins and skins


wincent:
Haven't done a release in the while, but while cutting [v4.14.1-liferay.19](https://github.com/liferay/liferay-ckeditor/releases/tag/v4.14.1-liferay.19) today I noticed that we have quite a few changes in the build output beyond what you would expect based on the only PR that we ([this one](https://github.com/liferay/liferay-ckeditor/pull/156)) merged since the last release.

Some examples: note how both [skins](https://github.com/liferay/liferay-ckeditor/commits/master/ckeditor/skins/moono) and [plugins](https://github.com/liferay/liferay-ckeditor/commits/master/ckeditor/plugins/scayt) change with almost every release.

I am not sure about why the skins change every time, but at least for plugins, you can see [we download whatever the latest version is during `./ck.sh setup`](https://github.com/liferay/liferay-ckeditor/blob/fe0e9f92e96f7bc9e8bce836abf53675eb39a097/ck.sh#L279-L280). This is kind of necessary: people _do_ have to "freshen" their repo before running `./ck.sh build`, but it definitely increases our risk profile — we might want to ship a one-line patch, but we actually ship that patch plus whatever-changed-in-plugins-and-skins-since-last-time as well.

So this ticket is just to flag the risk here and note that we might want to do something about it (like "freezing" these resources inside the repo instead of downloading them). I am pretty sure we can for those plugins, but the skins might be harder (if I had to guess, I'd bet that is happening as part of CKEditor's own build process, and that could be harder to change).

Labeling as "rfc" because it's not yet clear _what_ we will or should do about this.

julien:
Just adding some notes here.

I noticed that the [scayt](https://ckeditor.com/cke4/addon/scayt)  and [wsc](https://ckeditor.com/cke4/addon/wsc) plugins are downloaded each time we run:  [`./ck.sh setup`](https://github.com/liferay/liferay-ckeditor/blob/429dc8a7acd0bd9cde9e3854c4d5619b370ea8d2/ck.sh#L279-L280) and [`./ck.sh update`](https://github.com/liferay/liferay-ckeditor/blob/429dc8a7acd0bd9cde9e3854c4d5619b370ea8d2/ck.sh#L383-L384).

The [wsc](https://ckeditor.com/cke4/addon/wsc) plugin, will reach its **end of life** at the end of the year:
[...]

PR #159, opened 2021-03-22
fix: use CKEditor's `editable` to sync content between editor and dialogs

julien:
By using the `editable` function, we make sure the editor's content
is not lost when switching from wysiwyg to source mode, and
when opening the "preview" dialog.

**Before**

**After**

### DIFF
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -44,7 +44,7 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 
 			var editor = dialog.getParentEditor();
 
-			codeMirrorEditor.setValue(editor.getData(true));
+			codeMirrorEditor.setValue(editor.getData());
 
 			var preview = dialog
 				.getContentElement('main', 'preview')
--- a/plugins/codemirror/dialogs/codemirrordialog.js
+++ b/plugins/codemirror/dialogs/codemirrordialog.js
@@ -236,8 +236,9 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 
 			if (newData !== oldData) {
 				editor.setData(newData);
-				editor.setMode('wysiwyg');
 			}
+
+			editor.setMode('wysiwyg');
 		},
 
 		onShow: function () {
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -156,7 +156,7 @@
 		proto: {
 			setData: function (data) {
 				this.setValue(data);
-				this.value = 'ready';
+				this.status = 'ready';
 				this.editor.fire('dataReady');
 			},
 
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -16,8 +16,6 @@
 
 				contentsSpace.append(textarea);
 
-				editor.editable(new codeMirrorEditable(editor, textarea));
-
 				instance.codeMirrorEditor = CodeMirror.fromTextArea(
 					textarea.$,
 					{
@@ -28,6 +26,12 @@
 				);
 
 				var oldData = editor.getData(1);
+
+				var editable = editor.editable(
+					new codeMirrorEditable(editor, textarea)
+				);
+
+				editable.setData(oldData);
 
 				instance.codeMirrorEditor.setValue(oldData);
 


plugins/codemirror/plugin.js
```diff
@@ -51,11 +56,11 @@
 			});
 		},
 
-		_handleCodeMirrorChange: function (editor, oldData) {
+		_handleCodeMirrorChange: function (editor, editable, oldData) {
```
wincent, line 59: Can we get rid of the unused `editor` parameter now?

plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -44,7 +44,9 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 
 			var editor = dialog.getParentEditor();
 
-			codeMirrorEditor.setValue(editor.getData(true));
+			var editable = editor.editable();
+
+			codeMirrorEditor.setValue(editable.getData());
```
wincent, line 49: What's the difference between getting the data from the `editable` vs getting it from the `editor`?

(Just curious; I don't actually understand how any of this works.)

wincent (commented):
I am not really familiar with this code, so thanks for the screenshot showing that it works. One suggestion (about getting rid of an unused parameter) and one question (just for my own curiosity) inline, but no blocking feedback from me.

plugins/codemirror/plugin.js
julien: Yes, of course, I didn't pay attention to that.

plugins/codemirror/dialogs/codemirrordialog.js
julien: That's a good question which I've also been asking myself ... 

My quick answer would be: there should be no difference; **but** my longer answer would be: 
Given my doubts; it's probably a good opportunity to make sure I'm doing things correctly.

> (Just curious; I don't actually understand how any of this works.)

This  (`codemirror`) plugin, [adds a new "mode"](https://github.com/liferay/liferay-ckeditor/blob/2bdf5767b53df9f807fbc90e5b118091f3a6def7/plugins/codemirror/plugin.js#L10) to ckeditor,  almost exactly like the [`sourcearea`](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/plugins/sourcearea/plugin.js) plugin, the only difference is that we're using CodeMirror to make the "source code" editor look a bit nicer  that an simple `textarea`.

I first introduced the `editable` in [`cd36001`](https://github.com/liferay/liferay-ckeditor/commit/cd360019946b67e733d2046e6813df44040ef3af), because of [this issue](https://issues.liferay.com/browse/LPS-127948). Since our `codemirror` plugin is based on the [`sourcearea`](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/plugins/sourcearea/plugin.js) plugin (it's actually a mix of `sourcearea` and [`sourcedialog`](https://github.com/ckeditor/ckeditor4/tree/eb2d11644a796cd13cd047cd55541430359f9317/plugins/sourcedialog)), I thought it would make sense to "copy" as much of its behavior as possible **except** that I'm doing it wrong ...

For example: 

[This is](https://github.com/liferay/liferay-ckeditor/blob/2bdf5767b53df9f807fbc90e5b118091f3a6def7/plugins/codemirror/plugin.js#L160) incorrect: it should be `this.status` and not `this.value` (which might be causing this or other bugs)

[This](https://github.com/liferay/liferay-ckeditor/blob/2bdf5767b53df9f807fbc90e5b118091f3a6def7/plugins/codemirror/dialogs/codemirrordialog.js#L247-L254) shouldn't be needed because, when the dialog is opened we create a `textarea` (that we wrap with `CodeMirror`) and [set it's value to `editable.getData`](https://github.com/liferay/liferay-ckeditor/blob/2bdf5767b53df9f807fbc90e5b118091f3a6def7/plugins/codemirror/dialogs/codemirrordialog.js#L47-L49) (which should have the same value as `editor.getData()`) so there's no need to check again in the `onShow` method, at least that what I think (and also what is said in the [documentation](https://ckeditor.com/docs/ckeditor4/latest/guide/dev_savedata.html#retrieving-data-from-ckeditor-4))

All this to say, that I'm going to have another look at this, because even if having the `editable` makes sense in source mode (to fix the LPS mentioned above), calling [`editor.setData` should automatically call the editable's `setData` method](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/core/editable.js#L849-L851).

So what I think I should do is:
- Create the `editable` when the source more is registered
- When CodeMirror dispatches it's `change` event, call `editor.setData` with whatever value `codeMirrorEditor` has
- When opening the "preview" dialog make sure `editor.getData()` is up to date

I doubt this answered your question, but it clearly showed me I should re-think this.

wincent (approved):
Well, this certainly seems to be in the right language. `#Approved`

seems-legit-starbucks

julien:
@wincent thanks!

While I'm at it, since I'm going to prepare a release, I might have a look at #157 and see if I can do something about it (or at least think about what we should do)

PR #160, opened 2021-04-08
fix: initialize codemirror with the editor data and update it on data ready

NemethNorbert:
Hey,

[LPS-128735](https://issues.liferay.com/browse/LPS-128735),

These changes will solve the issue on 7.3.x and 7.2.x, on master the issue remains reproducible, however that is more than likely to be caused by data engine, and its still under investigation.

Please review the changes, thanks

/cc @antonio-ortega

### DIFF
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -49,6 +49,16 @@
 					instance._handleEditorResize.bind(instance)
 				);
 
+				editor.on('dataReady', function (event) {
+					var newData = event.data;
+
+					var oldData = instance.codeMirrorEditor.getValue();
+
+					if (newData && newData !== oldData) {
+						instance.codeMirrorEditor.setValue(newData);
+					}
+				});
+
 				editor.fire('ariaWidget', this);
 
 				callback();
@@ -156,8 +166,8 @@
 		proto: {
 			setData: function (data) {
 				this.setValue(data);
-				this.status = 'ready';
-				this.editor.fire('dataReady');
+				this.value = 'ready';
+				this.editor.fire('dataReady', data);
 			},
 
 			getData: function () {


antonio-ortega:
Hi @NemethNorbert ,

Can't we first fix the issue in data engine so we are sure this is properly fixing the original issue in master as well and we know we don't need any extra change?

Thanks.

NemethNorbert:
> Hi @NemethNorbert ,
> 
> Can't we first fix the issue in data engine so we are sure this is properly fixing the original issue in master as well and we know we don't need any extra change?
> 
> Thanks.

I really wanted to do that, however we did separate the issue within the source plugin and it clearly resolves the issue on or older branches except master. Based on that it seems safe to push this forward and handle the data-engine separately. I pushed this forward as the customer is waiting for 7weeks now, and it's getting urgent to provide some solution. If you don't think it can cause unwanted issues due to the inability to test on master properly, I'd like to move forward with this.

wincent:
> I really wanted to do that, however we did separate the issue within the source plugin and it clearly resolves the issue on or older branches except master. Based on that it seems safe to push this forward and handle the data-engine separately.

Do we have a ticket we can track for the data-engine changes? I'd hate to leave a loose end here that comes back to bite us later because we overlook it. Is it intended that LPS-128735 should remain open until `master` is fixed?

NemethNorbert:
> > I really wanted to do that, however we did separate the issue within the source plugin and it clearly resolves the issue on or older branches except master. Based on that it seems safe to push this forward and handle the data-engine separately.
> 
> Do we have a ticket we can track for the data-engine changes? I'd hate to leave a loose end here that comes back to bite us later because we overlook it. Is it intended that LPS-128735 should remain open until `master` is fixed?

There is no ticket currently opened for the data-engine issue, I will sort that out. My plan is to create a dependancy between the two tickets, we keep LPS-128735 open or under review as QA won't be able to approve it anyway, but we will be able to provide a hotfix with an updated ckeditor to our customers. I think it could be a silver lining if you all agree

plugins/codemirror/plugin.js
```diff
@@ -156,8 +166,8 @@
 		proto: {
 			setData: function (data) {
 				this.setValue(data);
-				this.status = 'ready';
-				this.editor.fire('dataReady');
+				this.value = 'ready';
```
julien, line 169: Why do we want to change this. (use `value` instead of `status`)

Our `codemirror` plugin is based on the `sourcearea` plugin from ckeditor, in which they use [`status`](https://github.com/ckeditor/ckeditor4/blob/7481c473b85b14f0bbfcd24fb5b368ebbc1fc3f9/plugins/sourcearea/plugin.js#L108) and not `value`.

Here we're creating a class based on [`CKEDITOR.editable`](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editable.html) (i.e. a "subclass"), which does not have a `value` property, so that's why I'm asking.

Looking at the rest of the changes, I see that [here](https://github.com/liferay/liferay-ckeditor/pull/160/files#diff-9799baf72c5fc1f956bcb1e148ef99f646a31d84ca3559c4c19760b379853cf4R55) we're using 

```js
var oldData = instance.codeMirrorEditor.getValue();
```

But that's not the related to the `editable` subclass, and CodeMirror (or a CodeMirror instance to be more precise) will always have a [`getValue`](https://codemirror.net/doc/manual.html#getValue) method, so @NemethNorbert could you explain why this change is needed when you get a chance?

julien: Hey @NemethNorbert,

I checked these changes in `master` to check both [LPS-128735](https://issues.liferay.com/browse/LPS-128735) as well as [LPS-127948](https://issues.liferay.com/browse/LPS-127948); because the last time we touched our `codemirror` plugin it was to fix that issue. 

Everything works as expected so that's good news (in `master`, we get the **Expected Result** described in  [LPS-128735](https://issues.liferay.com/browse/LPS-128735)).

That said, I'm still convinced that we should use `this.status = 'ready';` and not `this.value = 'ready'`.

You can read more about an `editable`'s status [here](https://ckeditor.com/docs/ckeditor4/latest/api/CKEDITOR_editable.html#property-status). 

Unless we set that `status`, its value is going to be `"unloaded"`, as you can guess from the following screenshot 



If this needs to be merged quickly to prepare a release to be able to send it to `7.2.x`, `7.3.x` and `master`, I can send a new PR (based on this one, with the change I propose).

What do you think?

julien:
Closing in favor of #161

PR #162, opened 2021-04-19
fix: LPS-127014 Add balloontoolbar missing style

edalgrin:
This PR add the missing styles for the balloontoolbar

**Before**

**After**


ambrinchaudhary (approved):

julien:
@edalgrin I edited the pull request description to include screenshot of how it looked before these changes and how it looks with these changes. 

The CSS changes look OK to me, would you like our resident CSS expert to have a quick look or do you think it's not necessary?

julien (approved):

edalgrin:
> would you like our resident CSS expert to have a quick look or do you think it's not necessary?

I would say it's not necessary, not because I think "it can't be wrong" but because it's harder to understand the context and why I'm changing that code than the code itself 😬

julien:
Since this only affects the `balloontoolbar` plugin when used with the `monoo-lexicon` skin: which isn't currently used anywhere there's very little of breaking anything, so I'm merging this. Thanks for your help @edalgrin

Issue #163, opened 2021-04-20
CKEditor is not displaying the full source code text on 7.0.x


georgel-pop-lr:
Steps to reproduce
1. Set CKEditor as your editor for Web Content
2. Start Liferay 7.0, fixpack-97 or higher
3. Access to create a new web content
4. Write a very long text with multiple new lines, so there is a scroll visible
5. Go to Source code

Expected behaviour
Source code text is completely visible

Actual behaviour
Source code text is not completely visible, we can't scroll until the end

For more details see: https://issues.liferay.com/browse/LPS-130834

PR #165, opened 2021-04-20
fix: LPS-130834 CKEditor is not displaying the full source code text on 7.0.x

georgel-pop-lr:
This PR fixes https://github.com/liferay/liferay-ckeditor/issues/163

### DIFF
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -35,6 +35,15 @@
 
 				instance.codeMirrorEditor.setValue(oldData);
 
+				var editableParent = editable.getParent();
+				var contentsSize = editableParent.getClientSize();
+				if (contentsSize.height) {
+					instance.codeMirrorEditor.setSize(
+						null,
+						contentsSize.height
+					);
+				}
+
 				instance.codeMirrorEditor.on(
 					'change',
 					instance._handleCodeMirrorChange.bind(
@@ -74,7 +83,7 @@
 		},
 
 		_handleEditorResize: function (event) {
-			this.codeMirrorEditor.setSize(null, event.data.outerHeight);
+			this.codeMirrorEditor.setSize(null, event.data.contentsHeight);
 		},
 
 		hidpi: true,


wincent:
@georgel-pop-lr: Are you able to [include a "Test plan"](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/pull_requests.md#prefer) here? Basically, want to see what steps you took to verify that this fixes the issue on 7.0, and doesn't regress anything on later branches.

julien:
@georgel-pop-lr just so you know; I tested this on `master`, following the test plan you left in the [original issue](#163) (*create a web content article with lots of text and switched to source mode*) and I didn't see anything strange.

georgel-pop-lr:
> @georgel-pop-lr just so you know; I tested this on `master`, following the test plan you left in the [original issue](#163) (_create a web content article with lots of text and switched to source mode_) and I didn't see anything strange.

Hi @julien,

It is true as this seems to affect only 7.0.x and would not happen to master as we have a scroll outside the editor and that's why this fix should not affect master. 

Please check on 7.0.x

Regards.

georgel-pop-lr:
> @georgel-pop-lr: Are you able to [include a "Test plan"](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/pull_requests.md#prefer) here? Basically, want to see what steps you took to verify that this fixes the issue on 7.0, and doesn't regress anything on later branches.

Hi @wincent,

I already had a call with @julien and explained why we need these changes, basically we need to synchronize the height of code mirror with the height of the editor. 

I changed the commit so is including the initialization of the height from config when we first open source code.

With these changes I confirm that everything works as expected.

Tomorrow I will check if any branches would be impacted by these, but it shouldn't.

Regards.

georgel-pop-lr:
> > @georgel-pop-lr: Are you able to [include a "Test plan"](https://github.com/liferay/liferay-frontend-projects/blob/master/guidelines/general/pull_requests.md#prefer) here? Basically, want to see what steps you took to verify that this fixes the issue on 7.0, and doesn't regress anything on later branches.
> 
> Hi @wincent,
> 
> I already had a call with @julien and explained why we need these changes, basically we need to synchronize the height of code mirror with the height of the editor.
> 
> I changed the commit so is including the initialization of the height from config when we first open source code.
> 
> With these changes I confirm that everything works as expected.
> 
> Tomorrow I will check if any branches would be impacted by these, but it shouldn't.
> 
> Regards.

Hi @wincent @julien,

I check on

plugins/codemirror/plugin.js
```diff
@@ -35,6 +35,15 @@
 
 				instance.codeMirrorEditor.setValue(oldData);
 
+				var editorElementParent = editor.element.getParent();
+				var contentsElement = editorElementParent.findOne(
+					'.cke_contents.cke_reset'
+				);
+				var contentsHeight = contentsElement.getClientSize().height;
+				if (contentsHeight) {
```
julien, line 43: Hey @georgel-pop-lr,

Since I saw these changes yesterday with you during our office hours session, I'm mostly OK with this.

I was just wondering why you added this `if (contentsHeight)` statement after using `contentHeight`.

Another thing I want to be sure about is that this also works in `master`; and I just want to have another look at this to make sure it also works outside DXP. So let me fetch this branch and see if there's not another "standard CKEditor way" to make this work.

Seems reasonable?

georgel-pop-lr: Hi @julien,

Yes seems reasonable to me, I did check the most I could but is always good to double-check as any change in ckeditor could impact many places.

I added this `if (contentsHeight)` because the height is getting applied from configuration, not sure if always is available, so I thought to add this condition. I did not check what happens if you don't apply any height on the config. 

Regards.

julien: @georgel-pop-lr,

Instead of 
```js
var editorElementParent = editor.element.getParent();
var contentsElement = editorElementParent.findOne(
    '.cke_contents.cke_reset'
);
var contentsHeight = contentsElement.getClientSize().height;
if (contentsHeight) {
    instance.codeMirrorEditor.setSize(null, contentsHeight);
}
```

How about 
```js
var editableParent = editable.getParent(); // should never be null 
var contentsSize = editableParent.getClientSize();
if (contentsSize.height) {
    instance.codeMirrorEditor.setSize(null, contentsSize.height);
}
```

What I like about this is that if ever ckeditor changes  it's class names (I doubt that would happen but just in case), this would still work. But I only tested on `master`.

What do you think?

georgel-pop-lr: Hey @julien,

Yes I like your proposal, I check on 7.0.x, and it works properly, best not to check by classes in case they change.

Regards.

Issue #166, opened 2021-05-06
removeListener() may be called on an undefined object


IstvanD:
Calling removeListener() may lead to an exception because the `me` object can be null/undefined.
https://github.com/ckeditor/ckeditor4/blob/major/core/event.js#L167

This may lead to issues like [LPS-131699](https://issues.liferay.com/browse/LPS-131699).

See [PTR-2398](https://issues.liferay.com/browse/PTR-2398) for more info and explanation.

A null check should be applied.

julien:
Fixed by #168

PR #168, opened 2021-05-07
fix: add null check

IstvanD:
Hello,

[LPS-131699](https://issues.liferay.com/browse/LPS-131699),

These changes will solve the issue on 7.3.x.
On master, the issue is not reproducible, because, on that branch, Web Content uses ckeditor4-react instead of liferay-ckeditor.

Please review the changes, thanks

### DIFF
--- a/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
+++ b/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
@@ -1,4 +1,4 @@
-From 5f6ea3154ececf568fed6168df109982723bfb4c Mon Sep 17 00:00:00 2001
+From 0daceabbc2fa55b8dc49c8e33355d9f15a9ac82e Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Tue, 14 May 2019 10:47:25 +0200
 Subject: [PATCH] LPS-89596 Cannot Drag Image from top content line in IE11
@@ -10,7 +10,7 @@ Changed position of drag icon for IE.
  1 file changed, 4 insertions(+)
 
 diff --git a/plugins/widget/plugin.js b/plugins/widget/plugin.js
-index 1d4e036fc..a53a4c809 100644
+index 8364a56e4..8cbc67e60 100644
 --- a/plugins/widget/plugin.js
 +++ b/plugins/widget/plugin.js
 @@ -1607,6 +1607,10 @@
--- a/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
+++ b/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
@@ -1,4 +1,4 @@
-From a94350e2a5272963fc9729b9dab1712dc1dadee5 Mon Sep 17 00:00:00 2001
+From e0e24be6d941b9696bf9b0fd25372b2a02014ac4 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Roland=20P=C3=A1kai?= <roland.pakai@liferay.com>
 Date: Tue, 21 May 2019 09:38:15 +0200
 Subject: [PATCH] LPS-95472 Tabs in popups not appears correctly in maximized
@@ -11,10 +11,10 @@ plugin.
  1 file changed, 1 insertion(+), 1 deletion(-)
 
 diff --git a/plugins/dialog/plugin.js b/plugins/dialog/plugin.js
-index 022d7b473..5537eef39 100644
+index 9a8d6fb20..8a94776f5 100644
 --- a/plugins/dialog/plugin.js
 +++ b/plugins/dialog/plugin.js
-@@ -1183,7 +1183,7 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
+@@ -1184,7 +1184,7 @@ CKEDITOR.DIALOG_STATE_BUSY = 2;
  					children: contents.elements,
  					expand: !!contents.expand,
  					padding: contents.padding,
--- a/patches/0003-LPS-85326-Remove-check-for-Webkit-browsers.patch
+++ b/patches/0003-LPS-85326-Remove-check-for-Webkit-browsers.patch
@@ -1,4 +1,4 @@
-From aa5d0c610dfc7131cf4405e706c215208774bc01 Mon Sep 17 00:00:00 2001
+From 0bf369f37c92ff1dda81f917fd564917de656f24 Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Wed, 25 Mar 2020 12:58:15 +0100
 Subject: [PATCH] LPS-85326 Remove check for Webkit browsers
@@ -18,7 +18,7 @@ https://github.com/liferay/liferay-ckeditor/commit/328017c65063ac18119e244fec92d
  1 file changed, 1 insertion(+), 1 deletion(-)
 
 diff --git a/core/selection.js b/core/selection.js
-index 0c0cb261c..dd7f4701e 100644
+index 61247b33f..bd05660fb 100644
 --- a/core/selection.js
 +++ b/core/selection.js
 @@ -2250,7 +2250,7 @@
--- a/patches/0004-LPP-36989-Remove-obsolete-summary-field-from-table-e.patch
+++ b/patches/0004-LPP-36989-Remove-obsolete-summary-field-from-table-e.patch
@@ -1,4 +1,4 @@
-From 754ecf387f47aaacd4265afd31ac148c6de2993f Mon Sep 17 00:00:00 2001
+From 7fc645c749ecc0554f5bc6f082254cbb93ed31db Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Roland=20P=C3=A1kai?= <roland.pakai@liferay.com>
 Date: Tue, 14 Apr 2020 10:15:56 +0200
 Subject: [PATCH] LPP-36989 Remove obsolete summary field from table elements
@@ -15,7 +15,7 @@ Subject: [PATCH] LPP-36989 Remove obsolete summary field from table elements
  8 files changed, 38 insertions(+), 56 deletions(-)
 
 diff --git a/plugins/table/dialogs/table.js b/plugins/table/dialogs/table.js
-index 99c90d7fa..c3b9531bd 100755
+index 5c42bc5c0..b19c2515d 100755
 --- a/plugins/table/dialogs/table.js
 +++ b/plugins/table/dialogs/table.js
 @@ -523,22 +523,6 @@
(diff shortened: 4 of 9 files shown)


NimrodPapp:
I'm not entirely sure who to mention, but based on previous PRs @antonio-ortega or @wincent, could you please review this PR for us?

Thank you,
Nimród

julien:
@IstvanD, thanks for submitting this.

I'm not sure this should be patched in CKEditor instead of being fixed in DXP:

I think that patching CKEditor's [event emitter](https://github.com/ckeditor/ckeditor4/blob/eb2d11644a796cd13cd047cd55541430359f9317/core/event.js#L149-L151) is taking a risk without really know what impact it could have. The patch isn't huge so you could think it's not a big deal, but my take on this is that if this was a CKEditor issue it would probably already have been fixed internally. Did you check the [issues section](https://github.com/ckeditor/ckeditor4/issues) on their GitHub to search for anything related?

> On master, the issue is not reproducible, because, on that branch, Web Content uses ckeditor4-react instead of liferay-ckeditor.

If I remember correctly, even though Web Content uses `ckeditor4-react`, we are actually loading the `ckeditor.js` file from the `liferay-ckeditor`package, this is visible [here](https://github.com/liferay/liferay-portal/blob/6bcd6f83bd8bcc185830b6d87d267a217df3c2e1/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/editor/Editor.js#L18) and [here](https://github.com/liferay/liferay-portal/blob/6bcd6f83bd8bcc185830b6d87d267a217df3c2e1/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/editor/Editor.js#L35-L36).

Additionally, in `master`, we've had a [very similar issue](https://issues.liferay.com/browse/LPS-129456).
 
You'll see that in the [event listener we use for the `resize` event](https://github.com/liferay/liferay-portal/blob/6bcd6f83bd8bcc185830b6d87d267a217df3c2e1/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/ckeditor.jsp#L592-L621), that the editor is being destroyed when the `window` is resized, maybe this would be a good place to start debugging.

Let me know what you think when you can.

julien (changes requested):
Requesting changes, in order to prevent an accidental merge.

antonio-ortega:
Hi guys,

Let me try to bring some more information to this topic.

@julien We've already followed your steps debugging from that point and we saw what you can read in [this comment](https://issues.liferay.com/browse/PTR-2398?focusedCommentId=2429383&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2429383).

@IstvanD When you submit a new pull request try to provide as much information as possible. Otherwise it is very difficult to understand for reviewers and prone to misunderstandings as you can see. Apart from that important point you should follow every time you submit a pull request, I'd like to ask you to clarify something else. In y[our comment](https://issues.liferay.com/browse/PTR-2398?focusedCommentId=2432790&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2432790) in previous PTR you said:

> On master, where the issue is not reproducible, ckeditor4-react is used instead of ckeditor4.
> There is a null check on prevHandler before removeListener is called.

I guessed you meant that null check is part of ckeditor4-react. Could you link that code, please?

And I understand, as well, you debugged it and check that null check in `ckeditor4-react` is called in Master opposite to what @julien said that we are using `ckeditor.js` from our internal `liferay-ckeditor` repo. @IstvanD Could you confirm you carried out those steps and you confirmed we are using `ckeditor4-react `code?

Thanks.

Regards.

IstvanD:
Hi guys,

In that comment, I meant this null-check code, but I forgot to link it.
https://github.com/ckeditor/ckeditor4-react/blob/master/src/ckeditor.jsx#L120

I debugged the issue in the following way on Liferay:

1. started creating a new web content
2. in the browser's JS debugger (F12), I searched for the {{ckeditor.js}} file
3. in {{ckeditor.js}}, I searched for the call of the {{removeListener()}} method. this is where my findings on 7.3.x and master differ, as removeListener() is not called the same way.

[7.3.x](https://github.com/ckeditor/ckeditor4/blob/major/core/event.js#L167)
[master](https://github.com/ckeditor/ckeditor4-react/blob/master/src/ckeditor.jsx#L121)

the main point is, that on master, a null-check for {{prevHandler}} preceedes the call of {{removeListener()}}.

Actually, I had no idea where this code resides
```
	_attachEventHandler( propName, prevHandler ) {
		const evtName = `${ propName[ 2 ].toLowerCase() }${ propName.substr( 3 ) }`;

		if ( prevHandler ) {
			this.editor.removeListener( evtName, prevHandler );
		}

		this.editor.on( evtName, this.props[ propName ] );
	}
```
as I have found it nowhere in the [ckeditor](https://github.com/ckeditor/ckeditor4) project.
What I did next will likely cause you to pull your hair out: I searched the code snippet on google. :)
I found the [ckeditor4-react](https://github.com/ckeditor/ckeditor4-react) repository, and there, I found [ckeditor.jsx](https://github.com/ckeditor/ckeditor4-react/blob/master/src/ckeditor.jsx) where I could locate the snippet.

I deduced, that on master, when creating a web content, {{ckeditor.jsx}} from ckeditor4-react get's activated instead of {{ckeditor.js}} from ckeditor, but I couldn't find the reason of this.

Best regards,
István

julien:
Hey @IstvanD,

Thanks for the information. I'll try to clear some doubts you might have, but my main point is that in [this](https://github.com/ckeditor/ckeditor4/blob/8b208d05d1338d046cdc8f971c9faf21604dd75d/core/event.js#L166-L168) function that we are patching:

```js
function removeListener() {
     me.removeListener( eventName, listenerFunction );
}
```

If  `me` is `undefined` or `null` we probably have a way of figuring out why by looking at the call stack.
CKEditor has [tests](https://github.com/ckeditor/ckeditor4/blob/8b208d05d1338d046cdc8f971c9faf21604dd75d/tests/core/event.js) to cover its event emitter and I'm *still* not convinced that we should patch CKEditor, but rather fix it on the DXP side.

I agree with you that the fix might end up being very similar to [this](https://github.com/ckeditor/ckeditor4-react/blob/c0c58bdb26d9b3cf401971e0cf4305a46a2a6fda/src/ckeditor.jsx#L121) (the link you posted earlier)

```js
if ( handler ) {
     editor.removeListener( evtName, handler );
}
```

The `CKEditor.jsx` (React) component just instantiates a CKEDITOR instance, just like we do in [here](https://github.com/liferay-frontend/liferay-portal/blob/6bcd6f83bd8bcc185830b6d87d267a217df3c2e1/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/ckeditor.jsp#L482). So maybe we need to do some checking on our side before calling the editor's `removeListener` function. 

> What I did next will likely cause you to pull your hair out: I searched the code snippet on google. :)

I actually think that being able to use google effectively is a skill, so no worries on my side :clap: 

> I deduced, that on master, when creating a web content, ckeditor.jsx from ckeditor4-react get's activated instead of ckeditor.js from ckeditor, but I couldn't find the reason for this.

I don't have all the details, but maybe a screenshot and some links might help:

This is Web Content in `master`,  in my browser with [React Developer Tools](https://chrome.google.com/webstore/detail/react-developer-tools/fmkadmapgofadopljbjfkapdkoienihi?hl=en) extension installed:



You'll see that I've highlighted a `RichText` component which in turn renders various children, among which is our (beloved) `CKEditor` component. When looking at [this](https://github.com/liferay-frontend/liferay-portal/blob/master/modules/apps/dynamic-data-mapping/dynamic-data-mapping-form-field-type/src/main/resources/META-INF/resources/RichText/RichText.es.js#L63) line I think that it should make even more sense. 

To go a bit deeper down the rabbit hole, I checked what was going on in the `journal-web` module (Web Content), and this is what I see is that this `RichText` component is probably rendered by the `<liferay-data-engine:data-layout-renderer />` tag [here](https://github.com/liferay/liferay-portal/blob/6bcd6f83bd8bcc185830b6d87d267a217df3c2e1/modules/apps/journal/journal-web/src/main/resources/META-INF/resources/edit_article.jsp#L245-L253). Now I could go deeper and also check the source code of that tag but that still wouldn't solve our problem. 
On `7.3.x` this is [slightly different](https://github.com/liferay/liferay-portal/blob/3cc350081830d5b3ed7848d769d3985a6bbf0469/modules/apps/journal/journal-web/src/main/resources/META-INF/resources/edit_article.jsp#L265-L279) and I'm assuming that the `<liferay-ddm:html />` tag renders `ckeditor.jsp` at some point.

I hope that helps, but let me know otherwise.

@antonio-ortega I read your comment (thanks for the extra info), and what I get from it is that: 

> we are not clearing all listeners as expected when changing locale

and that's what should focus on fixing. Is that correct?

If you guys need a hand with this, please let me know.

IstvanD:
Hi @julien 

Thank you very much for all the explanation. I think I understand much better why the differences occur on these different branches.

But I'm still having a hard time figuring out how to solve the issue on the DXP side.
What is certain is that it occurs only after the backport of LPS-129456 and the backport version is identical to master.
For some reason, adding the `onInitEditor()` callback causes `me` not to be left undefined for some occasions, but I couldn't find the connection reason despite exhaustive debugging.

Best regards,
István

julien:
@IstvanD no problem, always happy to help.
I'll take a look at this during the afternoon and tell you if I can spot anything.

IstvanD:
Hi @julien I just wanted to catch up on this topic.
Did you manage to find anything on the DXP side that may have led to the exception?

Thank you, and best regards,
István

julien:
Hi @IstvanD sorry for the late reply.

I finally managed to have a look into this and I don't think the issue is going to be trivial to fix.

The first "fishy" thing I'm seeing when we resize the browser's viewport is this message in the console.

liferay-component-warning

I'm not saying it's related to our "duplicated paste" problem, but it's not a good sign, and should be fixed.

I also think that this issue has something to do with event listeners, in the sense that I have a feeling that one or more event listeners are not getting "removed" when we switch from one language to another. Then again, given all the different event handlers that manipulate ckeditor data, I'm haven't been able to spot which one was causing this bug (it's a bit like looking for a needle in a haystack).

The next thing I tried was fetching these changes, building `liferay-ckeditor` in DEBUG mode and linking it in dxp, in order to see if that fixed the issue, and **it does**. 

So given that this is a severe bug, I have finally changed my mind about accepting this solution.

Unfortunately, there is a *slight* problem with this pull request because all the previous patches [have been deleted](https://github.com/liferay/liferay-ckeditor/pull/168/files), so could you please resend it (I don't really mind if you push force or send a new one). 

Thanks!

julien:
I also forgot about @antonio-ortega 's [comment](https://issues.liferay.com/browse/PTR-2398?focusedCommentId=2429383&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2429383) on the Jira ticket.

Taking that into account, I debugged [that part](https://github.com/ckeditor/ckeditor4/blob/a8d566a548ae2f87d833f2a40390355263326d07/core/editable.js#L86-L89) of the code (and added some "log" points) and this is what I see, when I switch from one language to another 



And here's the same thing with a bit more details



What's strange at first sight is that the `listener` doesn't really seem `undefined` (it just looks like an object literal) but Chrome's devtools show that it has a `me` property in it's "[[ Scopes ]]" (no idea what that actually is but I'll search :shrug:) that is `undefined`. So maybe some weird race condition going on? 

It also seems that this is coming from the `clipboard` plugin, because that's where the [`initDragDataTransfer`](https://github.com/ckeditor/ckeditor4/blob/8b208d05d1338d046cdc8f971c9faf21604dd75d/plugins/clipboard/plugin.js#L2197) function is defined ... although other listeners also showed up :partying_face: 

All that to say, that this patch doesn't seem wrong to me after all.

julien (approved):

julien:
Fixes #166

PR #174, opened 2021-05-13
chore: update @clayui/css dependency to v3.27.0

julien:
In order to make sure have the latest icons available in `@clayui/css`
because we're going to need them for [LPS-132048](https://issues.liferay.com/browse/LPS-132048)

And because of this we'll also need to cut a release :tada:


julien:
Closing in favour of #175

PR #175, opened 2021-05-13
fix: update @clayui/css to latest and add more icons

julien:
In order to make sure have the latest icons avaible in `@clayui/css`
because we're going to need them for [LPS-132048](https://issues.liferay.com/browse/LPS-132048)

Additionally as part of [LPS-127012](https://issues.liferay.com/browse/LPS-127012),
we're adding more icons to our `moono-lexicon` skin that will be used in
the Balloon Editor

### DIFF
--- a/skins/moono-lexicon/icons.json
+++ b/skins/moono-lexicon/icons.json
@@ -1,6 +1,9 @@
 {
     "dir": "../../node_modules/@clayui/css/lib/images/icons",
     "icons": {
+        "add-cell": "add-cell",
+        "add-column": "add-column",
+        "add-row": "add-row",
         "align-image-center": "align-image-center",
         "align-image-left": "align-image-left",
         "align-image-right": "align-image-right",
@@ -16,6 +19,8 @@
             "ltr": "list-ul",
             "rtl": "list-ul-rtl"
         },
+        "check": "check",
+        "color-picker": "color-picker",
         "copy": {
             "ltr": "copy",
             "rtl": "copy"
@@ -28,8 +33,10 @@
             "ltr": "search",
             "rtl": "search"
         },
+        "folder": "folder",
         "hidden": "hidden",
         "horizontalrule": "separator",
+        "hr": "hr",
         "image": "picture",
         "imageselector": "picture",
         "indent": {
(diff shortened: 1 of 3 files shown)


skins/moono-lexicon/icons.json
```diff
@@ -16,6 +19,8 @@
             "ltr": "list-ul",
             "rtl": "list-ul-rtl"
         },
+        "check": "check",
+        "color-picker": "color-picker",
```
carloslancha, line 23: I may be wrong but this icon looks like one that should have a rtl version

skins/moono-lexicon/icons.json
```diff
@@ -28,8 +33,10 @@
             "ltr": "search",
             "rtl": "search"
         },
+        "folder": "folder",
```
carloslancha, line 36: I may be wrong but this icon looks like one that should have a rtl version

skins/moono-lexicon/icons.json
```diff
@@ -42,6 +49,7 @@
         "justifyleft": "align-left",
         "justifyright": "align-right",
         "link": "link",
+        "low-vision": "low-vision",
```
carloslancha, line 52: I may be wrong but this icon looks like one that should have a rtl version

carloslancha (commented):
Hey @julien could you please check my comments with @drakonux ? Thx! 😄

julien:
> Hey @julien could you please check my comments with @drakonux ? Thx! smile

Of course, will do. Your comments make sense though, but I didn't find those rtl versions (maybe that I didn't check properly).

drakonux:
The check, folder, and color-picker icons don't need to be mirrored because are common symbols where the direction is not relevant (e.g. message icon has to be mirrored because the direction is relevant in an RTL scenario). 
The low vision icon doesn't need to be mirrored because slashes are not mirrored either.

julien:
Thanks for the quick confirmation @drakonux

julien:
In any case and just to keep the conversation going: using CSS transform, would probably be enough to get a missing rtl icon. Doesn't sound like a crazy idea.

drakonux:
Yeah, in fact probably I created some RTL icons that I shouldn't... so little by little, we will adjust this because I'm a bit inexperienced in this topic and I think none of us.

markocikos:
LGTM!

PR #176, opened 2021-05-20
fix: remove unneeded styles

julien:
This fixes the visual glitch produced when selecting an element
in the CKEDITOR.ui.richCombo element used by the [Styles
Combo](https://ckeditor.com/cke4/addon/stylescombo) plugin

**Before**

**After**

**Warning**
Depends on #175 

[EDIT] If you think it's a better idea to fold these changes into #175, I can do it too.


markocikos:
LGTM!

PR #179, opened 2021-06-02
chore: update ckeditor-dev to 4.16.1

jonmak08:
Hey @liferay,

Attached is an update for to upgrade CKEditor to [4.16.1](https://github.com/ckeditor/ckeditor4/tree/4.16.1).

Please let me know if you have any questions. Thanks!


julien:
Hi @jonmak08, 
The code changes look OK to me. We usually send a pull request updating ckeditor to the new version and an automatic commit is added when running the build(+applying patches) and releasing a new version, but I don't think it's much of  a problem to do it this way either.
I've also tested this in DXP and haven't seen any problem so you've got my approval.

julien (approved):

markocikos (approved):
LGTM!

jonmak08:
Thanks guys!  I was wondering what the workflow was, but I figured I would send the base changes over first then send all the merge commits and build changes next since those seem to occur _after_ the merge commit based on the Git history I see.

Thanks for reviewing guys!

PR #183, opened 2021-06-28
fix: retrieve hspace and vspace values from style

### LINKED ISSUE #182: CKEditor image-properties vspace and hspace values not retrieved
**Issue description**: CKEditor image-properties vspace and hspace values not retrieved

**Reproduction steps**
1. Start Master and set ckeditor in portal-ext.properties ( editor.wysiwyg.portal-impl.portlet.ddm.text_html.ftl=ckeditor)
2. Go to Content & Data > Web content
3. Create a new Basic Web Content and add a new image in wysiwyg mode
4. Right click on the image and open image properties
5. Set a vspace and hspace value and save it
6. Switch to source mode, then back to wysiwyg mode and open the image properties

**Expected behavior:** when switching back from source to wysiwyg mode, hspace and vspace fields of image properties are filled with the value that was just saved in wysiwyg mode
**Actual behavior:** when switching back from source to wysiwyg mode, hspace and vspace fields of image properties are not filled with the value that was just saved in wysiwyg mode

Root cause analysis and more information on [PTR-2511](https://issues.liferay.com/browse/PTR-2511)

**Solution proposal:** add onShow function where we extract the widget from editor and interpret the styles applied to it with regular expression

georgel-pop-lr:
Please see [LPS-134777](https://issues.liferay.com/browse/LPS-134777) for more information.

We could have multiple cases scenarios, take in mind that a user can change the source code as he likes, so we need to check on each dialogue display the styles of the selected image.
These are all the possible cases I found, if the user did:
1. not apply any style or hspace or vspace in the dialog, this will mean empty values

2. apply only hspace, the style would be: style="margin-left: 10px; margin-right: 10px;"

3. apply only one vspace, the style would be: style="margin-bottom: 10px; margin-top: 10px;"

4. apply hspace and vspace, the style would be: style="margin: 20px 10px;"

5. applied the same value for hspace and vspace, the style would be: style="margin: 10px;"

6. change this style from source code as he wants, so we need to interpret this values as possible as we can, that why I put all the conditions above.

/cc @orsolyaDekany 

Fixes #182

### DIFF
--- a/plugins/imagespacingbox/plugin.js
+++ b/plugins/imagespacingbox/plugin.js
@@ -19,6 +19,36 @@
 							{
 								children: [
 									{
+										onShow: function () {
+											var parentEditor = this.getDialog().getParentEditor();
+											var widget =
+												parentEditor.widgets.focused;
+
+											if (widget) {
+												var imageElement =
+													widget.parts.image.$;
+												var style = imageElement.style;
+
+												var hspace = '';
+
+												if (style) {
+													hspace =
+														parseInt(
+															style.marginLeft,
+															10
+														) ||
+														parseInt(
+															style.marginRight,
+															10
+														);
+												}
+
+												widget.setData(
+													'hspace',
+													hspace
+												);
+											}
+										},
 										commit: function (widget) {
 											widget.setData(
 												'hspace',
@@ -57,6 +87,36 @@
 							{
 								children: [
 									{
+										onShow: function () {
+											var parentEditor = this.getDialog().getParentEditor();
+											var widget =
+												parentEditor.widgets.focused;
+
+											if (widget) {
+												var imageElement =
+													widget.parts.image.$;
+												var style = imageElement.style;
+
+												var vspace = '';
+
+												if (style) {
+													vspace =
+														parseInt(
+															style.marginTop,
+															10
+														) ||
+														parseInt(
+															style.marginBottom,
+															10
+														);
+												}
+
+												widget.setData(
+													'vspace',
+													vspace
+												);
+											}
+										},
 										commit: function (widget) {
 											widget.setData(
 												'vspace',
--- a/plugins/imagespacingbox/plugin.js
+++ b/plugins/imagespacingbox/plugin.js
@@ -55,12 +55,14 @@
 												this.getValue()
 											);
 
-											var hspace = widget.data.hspace;
+											var hspace = widget.data.hspace
+												? widget.data.hspace
+												: 0;
 
 											var imageElement =
 												widget.parts.image;
 
-											if (imageElement && hspace) {
+											if (imageElement) {
 												imageElement.setStyles({
 													'margin-left':
 														hspace + 'px',
@@ -123,12 +125,14 @@
 												this.getValue()
 											);
 
-											var vspace = widget.data.vspace;
+											var vspace = widget.data.vspace
+												? widget.data.vspace
+												: 0;
 
 											var imageElement =
 												widget.parts.image;
 
-											if (imageElement && vspace) {
+											if (imageElement) {
 												imageElement.setStyles({
 													'margin-bottom':
 														vspace + 'px',


plugins/imagespacingbox/plugin.js
```diff
@@ -14,11 +14,89 @@
...
+
+						if (widget && widget.name == 'image') return widget;
+
+						return null;
+					};
+
+					var getImageHVSpace = function (widget) {
+						var imageHVSpace = {hspace: '', vspace: ''};
+						var imageElement = widget.parts.image;
+						var style = imageElement.getAttribute('style');
```
julien, line 28: Hey @georgel-pop-lr,

Instead of using `getAttribute` here, don't you think we could access the `CSSStyleDeclaration` directly to avoid using regular expressions?

```js
var imageElement = widget.parts.image.$; // Get the native DOM element (not the `CKEDITOR.dom.element`)
var style = imageElement.style; // This should return a CSSStyleDeclaration 
```

georgel-pop-lr: Hi @julien,

Thanks for reviewing, your idea is good, but I don't understand how we could remove the use of  regular expressions, because I use them to interpret and extract the value of the margin, in the style you could have others styles as well.

Regards.

julien: Well, if you have a `CSSStyleDeclaration` instance, you could use its `marginTop`, `marginRight`, `marginBottom` and `marginLeft` properties

georgel-pop-lr: Didn't know about CSSStyleDeclaration, makes total, sense, changes pushed. Please check.

Regards.

plugins/imagespacingbox/plugin.js
```diff
@@ -14,23 +14,65 @@
 						return;
 					}
 
+					var getFocusedWidget = function (editor) {
+						var widget = editor.widgets.focused;
+
+						if (widget && widget.name == 'image') return widget;
```
julien, line 20: I'm not sure this should be a function: I know it's been called in two places, but since we're returning `null` we could probably inline this and just check for `editor.widgets.focused`

If we really want to keep this as a function though, we should

1. Use strict equality comparison here (`===` and not `==`) 
2. Add some curly braces on that `if` statement

I know these aren't required to make it work, but since we apply those rules in all our frontend related code, I think it makes sense to apply them here too

plugins/imagespacingbox/plugin.js
```diff
@@ -14,23 +14,65 @@
...
+
+						return null;
+					};
+
+					var getImageHVSpace = function (widget) {
+						var imageHVSpace = {hspace: '', vspace: ''};
```
julien, line 26: Why are you making this an "internal" function, couldn't this be declared directly on the plugin?

julien:
@georgel-pop-lr I tested these changes and could see them working.

I left some comments for minor improvements. On top of that, I think adding a screenshot of the functionality *before* and *after* the changes would have made it easier for others to review.

You should edit the title to make it shorter (this isn't going to look pretty when doing `git log` or when generating the changelog), and the title shouldn't contain the LPS/Jira ticket reference, you can put that in the commit message.

julien (changes requested):
Requesting changes to avoid accidental merge

georgel-pop-lr:
> Requesting changes to avoid accidental merge

Changes done. Please review.

Regards.

julien (approved):
LGTM

julien:
@georgel-pop-lr I've edited the title and message before merging - thanks again.

rolandpakai:
Hi @julien 

where can i follow the v4.16.1-liferay.1 release in Liferay?

Thanks, Roland

julien:
Hi @rolandpakai,

I don't know, it should be done by the dxp squad.
Maybe @markocikos, @diegonvs, @carloslancha know

rolandpakai:
Hi @julien 

it seems it has been already relealsed: https://github.com/liferay/liferay-portal/commit/94020f9c55c4e4f1410e14124a43b4925a28caa2

But the v4.16.1-liferay.1 release not contains this PR:
https://github.com/liferay/liferay-ckeditor/compare/v4.16.0-liferay.2...v4.16.1-liferay.1

I would like to ask when will be released this changes?

Thanks, Roland

diegonvs:
Hey @rolandpakai and @georgel-pop-lr, [I released a new version of liferay-ckeditor containing this fix](https://github.com/liferay/liferay-ckeditor/blob/master/CHANGELOG.md#v4161-liferay2-2021-07-05) and I'll create a PR tomorrow(6 Jul).  Just FYI :D

rolandpakai:
Thank you @diegonvs

PR #187, opened 2021-08-06
fix: Maximize plugin breaks the UI in Firefox

### LINKED ISSUE #186: Maximize plugin breaks the UI in Firefox
This is caused by solution of this issue https://dev.ckeditor.com/ticket/5560
As it add contentEditable=true to the body of the html. This renders our UI in liferay unusable. Please see https://issues.liferay.com/browse/LPS-136998

Removing the changes introduced for 5560, won't reintroduce the described problem.

NemethNorbert:
This issue is we are putting contentEditable attribute to the body html, however it breaks the UI on Firefox. This change was introduce in https://dev.ckeditor.com/ticket/5560, reverting it does not reintroduce the issue.

Please review my changes,
Thanks! 

Fixes #186

### DIFF
--- a/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
+++ b/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
@@ -1,4 +1,4 @@
-From 69dba89b93f8f3873f24a182729f5db5edba0436 Mon Sep 17 00:00:00 2001
+From befb29f639df8edaa059c24bc540c7453eb68159 Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Tue, 14 May 2019 10:47:25 +0200
 Subject: [PATCH] LPS-89596 Cannot Drag Image from top content line in IE11
@@ -10,7 +10,7 @@ Changed position of drag icon for IE.
  1 file changed, 4 insertions(+)
 
 diff --git a/plugins/widget/plugin.js b/plugins/widget/plugin.js
-index 85b6103b17..47ebe4b012 100644
+index 85b6103b1..47ebe4b01 100644
 --- a/plugins/widget/plugin.js
 +++ b/plugins/widget/plugin.js
 @@ -1607,6 +1607,10 @@
--- a/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
+++ b/patches/0002-LPS-95472-Tabs-in-popups-not-appears-correctly-in-ma.patch
@@ -1,4 +1,4 @@
-From 9736aabd532e839c72f14e8038682081e4f2c05c Mon Sep 17 00:00:00 2001
+From 776bd4cd69d11597b5960909930473765ee923e5 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Roland=20P=C3=A1kai?= <roland.pakai@liferay.com>
 Date: Tue, 21 May 2019 09:38:15 +0200
 Subject: [PATCH] LPS-95472 Tabs in popups not appears correctly in maximized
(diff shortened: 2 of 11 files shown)


markocikos:
@NemethNorbert I couldn't reproduce the issue on DXP master. Looking at the [JIRA ticket](https://issues.liferay.com/browse/LPS-136998), it's not even reproducible on 7.3.x, but on 7.2.x-private. I don't think we should merge and maintain a patch that fixes a bug that was fixed a long time ago. 

You mentioned in [this JIRA comment](https://issues.liferay.com/browse/LPP-42246?focusedCommentId=2497025&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2497025) that the reason might be React implementation, but [ckeditor4-react](https://github.com/ckeditor/ckeditor4-react/blob/v1.4.2/src/ckeditor.jsx) in only a relatively small wrapper around base ckeditor, that enables it to be invoked as a React component. Nowhere in it is any mention of `contentEditable`. It is more likely that our update of base ckeditor version to 4.15.x and later 4.16.x fixed the issue. If a customer has this issue, they might try updating "liferay-ckeditor" to the latest "4.16.1-liferay.3", or some previous version that fixes the issue but has the least side-effects.

NemethNorbert:
Hey @markocikos ,

You are right, I updated the base ckeditor version to 4.16.1 as well as used the lates liferay-ckeditor version on 7.2.x, however the issue still persisted.
on 7.2.x `one.document.getBody();` will return the body of the page instead of the ckeditor iframe's body. On 7.2.x, we are missing  [LPS-110011](https://issues.liferay.com/browse/LPS-110011) [LPS-118027](https://issues.liferay.com/browse/LPS-118027) that turned the descriptionXML editor into a richtext editor by replacing the stripped down alloyeditor to ckeditor. Without those changes the summary field, does not contain the iframe and alloyeditor is used.

If we backport those tickets the issue is solved on 7.2.x as well, however it changes the summary field functionality and presentation. Are we good to make that change on 7.2.x ?

Thanks,

markocikos:
> Are we good to make that change on 7.2.x ?

LPS-110011 and LPS-118027 seem to be significant changes, new features. I am not sure if we should backport these. Normally we don't do backports of larger features, as it might introduce issues in the old environment and prove to be costly to fix. In contrast, the issue of this PR is a pretty minor one: it is a firefox-only bug, on a plugin we don't include by default. A known solution is simply to upgrade to 7.3 DXP. I would honestly recommend that we 'close / won't fix' this. If we need to be very diplomatic with this customer, maybe we can offer them a hotfix with packported LPS-110011 and LPS-118027? @nhpatt what do you think?

nhpatt:
LPS-110011 is unbackporteable to 7.2.x, it changes the behaviour for clients way after than the official release/first GAs. I don't think they should be backported.

> A known solution is simply to upgrade to 7.3 DXP. I would honestly recommend that we 'close / won't fix' this. 

This puts all the burden in the client. I don't know how important is this for them and if we can just accept it and tell them that is already fixed in newer versions. If it's a stopper I would merge the patch instead of backporting for all the users.

> Maybe we can offer this customer a hotfix with backported LPS-110011 and LPS-118027? @nhpatt what do you think?

That would be great but I don't think they do hotfixes based on things that are not integrated in 7.2.x, from branches, just for one client. If that possibility exists, patching that just for them I think it would be the best solution.

markocikos:
Make sense. I'd say the way forward is to check with the customer:
1. How important is this fix?
2. Do they have plans to upgrade to 7.3?

If it's important, and they don't have plans to upgrade, then we merge this PR and publish.

markocikos:
@NemethNorbert It is very valuable for us to the least possible amount of customizations of the base ckeditor. This enables us to easily upgrade base version, leveraging improvements from core ckeditor team. Every custom patch adds a possible future conflict or bug, and should be carefully considered before merging.

Having that in mind, it's not the end of the world to make a new patch, if it makes sense to do :slightly_smiling_face:

NemethNorbert:
Hey @markocikos,

We asked about the future plans to upgrade to 7.3, however we also just received another ticket regarding the same issue on Firefox with the maximise plugin from another customer. Because of that, I would lean towards merging this fix.

Thanks,

markocikos:
@NemethNorbert OK, let's merge this! Please rebase and rebuild, to fix conflict.

NemethNorbert:
Hey @markocikos ,

I updated the PR and resolved the conflict.

Thanks,

markocikos:
@NemethNorbert Please rebase your patch on top. It should be the 11th patch, not replace the 10th.

markocikos (approved):

PR #188, opened 2021-08-09
fix: LPS-136119 We can't open panel elements when the table toolbar is re-rendered

markocikos:
Fixes: https://issues.liferay.com/browse/LPS-136119

To reproduce:
1. Place `frontend-editor-ckeditor-sample-web` on a page.
2. See movies of the bug the linked ticket.
3. Comment out [this block of code](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins/ballooneditor/plugin.js#L387-L397), overriding of `CKEDITOR.ui.panel.prototype.showBlock`. This block is inadequate to fix this issue, and prevents the solution in this PR to work correctly.

The root cause of the issue is that `id` of buttons is changing [with each render](https://github.com/ckeditor/ckeditor4/blob/major/plugins/button/plugin.js#L146). On the other side, blocks are tied to `id` of the button that created them. `addBlock` sets `this._.blocks[ name ]` [only on creation](https://github.com/ckeditor/ckeditor4/blob/major/plugins/panel/plugin.js#L205). With the seconds render, link through `id` is lost.

Notes:
- This should not affect classic ckeditor, as buttons are rendered only once.
- `richcombo` is an example where `id` of trigger is set only once, [in constructor](https://github.com/ckeditor/ckeditor4/blob/cae20318d46745cc46c811da4e7d68b38ca32449/plugins/richcombo/plugin.js#L89), and not on each render.

After PR:
after

### DIFF
--- a/patches/0010-LPS-136119-Set-id-on-first-render-instead-of-changin.patch
+++ b/patches/0010-LPS-136119-Set-id-on-first-render-instead-of-changin.patch
@@ -1,0 +1,37 @@
+From a2ebadce31c58f4a725f95905ef97b6a4146d37f Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Marko=20=C4=8Ciko=C5=A1?= <marko.cikos@liferay.com>
+Date: Mon, 9 Aug 2021 18:04:44 +0200
+Subject: [PATCH] LPS-136119 Set `id` on first render, instead of changing it
+ with each render
+
+---
+ plugins/button/plugin.js | 9 ++++++++-
+ 1 file changed, 8 insertions(+), 1 deletion(-)
+
+diff --git a/plugins/button/plugin.js b/plugins/button/plugin.js
+index 9cc0584169..e0c996e9e4 100644
+--- a/plugins/button/plugin.js
++++ b/plugins/button/plugin.js
+@@ -143,7 +143,6 @@
+ 			}
+ 
+ 			var env = CKEDITOR.env,
+-				id = this._.id = CKEDITOR.tools.getNextId(),
+ 				stateName = '',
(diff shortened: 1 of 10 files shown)


patches/0010-LPS-136119-Set-id-on-first-render-instead-of-changin.patch
```diff
@@ -0,0 +1,37 @@
...
+ 			var env = CKEDITOR.env,
+-				id = this._.id = CKEDITOR.tools.getNextId(),
+ 				stateName = '',
+ 				command = this.command,
+ 				// Get the command name.
+@@ -151,6 +150,14 @@
+ 				keystroke,
+ 				shortcut;
+ 
++			var id = this._.id;
```
nhpatt, line 27: Solution seems to be pretty simple, so LGTM.

nhpatt (approved):

diegonvs (approved):
Moving a comment from Slack to GH 😅 

I tested removing menubutton implementation and using our current implementation which is relying in panelbutton and it worked well! Thanks @markocikos :)))

I tested BGColor and text formatting button and they still working fine. Stressing with mounting and unmounting the panels and everything looks okay.

patches/0010-LPS-136119-Set-id-on-first-render-instead-of-changin.patch
diegonvs: I think you can try sending this solution to solve [this](https://github.com/ckeditor/ckeditor4/issues/4639) opened ticket in CKEditor's repo too

PR #189, opened 2021-08-17
fix: LPS-137425 Focusing on balloon editor toolbar may open an additional text toolbar

markocikos:
Fixes: https://issues.liferay.com/browse/LPS-137425

In this PR, we are removing force selection update on `focus`. Selection update has a lot of side-effects, including opening a balloon toolbar. I tried multiple times to makes a workaround, see https://github.com/liferay-frontend/liferay-portal/pull/1284 and https://github.com/liferay-frontend/liferay-portal/pull/1350, but each solution leads to a new bug. I finally ran out of options, the only solution remaining to fix the root cause. 

I was hesitant to do modify base ckeditor because the comment references `previousActive` sync. Looking in more depth in the ckeditor code, I think removing this block is safe to do:
- I believe [this block](https://github.com/ckeditor/ckeditor4/blob/cae20318d46745cc46c811da4e7d68b38ca32449/core/selection.js#L800-L815) achieves the intended sync,  If this is true, we are just cleaning up old code. 
- I tested blur/focus selection behavior on FF and Chrome, before and after this PR, and I didn't see any difference. 
- The purpose of the removed code does not seem critical, restoring selection on blur/focus. The worst case scenario is we re-introduce a minor bug, affecting an edge case. A good trade-off for proper functioning of balloon editor. If that's the case, we will want to add a more precise solution, instead of forcing selection that has many side-effects.

Before PR:
before

After PR:
after

### DIFF
--- a/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
+++ b/patches/0001-LPS-89596-Cannot-Drag-Image-from-top-content-line-in.patch
@@ -1,4 +1,4 @@
-From befb29f639df8edaa059c24bc540c7453eb68159 Mon Sep 17 00:00:00 2001
+From 3f85bfae11dd9c95ecdd4b43e4d99491fbbadd67 Mon Sep 17 00:00:00 2001
 From: Julien Castelain <julien.castelain@liferay.com>
 Date: Tue, 14 May 2019 10:47:25 +0200
 Subject: [PATCH] LPS-89596 Cannot Drag Image from top content line in IE11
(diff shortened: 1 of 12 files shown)


diegonvs:
> If that's the case, we will want to add a more precise solution, instead of forcing selection that has many side-effects.

I tried to see if we can `event.stopPropagation()` when clicking in linktoolbar's buttons but with the pattern `"onclick="CKEDITOR.tools.callFunction"` we don't receive event information. 😞 

In overall, considering _"The purpose of the removed code does not seem critical, restoring selection on blur/focus. The worst case scenario is we re-introduce a minor bug, affecting an edge case."_ it looks good to me.

PR #191, opened 2021-09-29
fix: [TS] LPS-139565

holatuwol:
https://issues.liferay.com/browse/LPS-139565

## Problem overview

In earlier versions of CKEditor, it set the height and width of a resized image as style properties. However, in later versions of CKEditor, it instead sets element attributes instead. As a result, any content created with the earlier version of CKEditor will run into issues in the newer versions of CKEditor.

## Steps to reproduce (in Liferay)

1. Navigate to the screen to create a new web content
2. Access the WYSIWYG editor and upload an image to the document library and add it to the content
3. Switch to the source view and add the `style="height:100px; width:100px"` attribute to the `img` tag to simulate content created in earlier versions of Liferay
4. Publish the web content
5. Edit the web content that you created
6. Attempt to resize the image using either the dragresizer or by double-clicking on the image to load Image Properties

Expected behavior is that reloading the content in WYSIWYG view will result in an image that can be resized. Actual behavior is that the inline styling makes it impossible to resize the image.

### DIFF
--- a/patches/0013-LPS-139565-When-upgrading-from-6.2-to-7.1-image-widt.patch
+++ b/patches/0013-LPS-139565-When-upgrading-from-6.2-to-7.1-image-widt.patch
@@ -1,0 +1,52 @@
+From 3edd637a32117992665065e07abbad4d556ca742 Mon Sep 17 00:00:00 2001
+From: Minhchau <minhchau.dang@liferay.com>
+Date: Tue, 28 Sep 2021 11:18:40 -0700
+Subject: [PATCH] LPS-139565 When upgrading from 6.2 to 7.1, image width/height
+ in web content is lost
+
+If inline styles for height/width are set on the image, prefer those over the natural height/width
+---
+ plugins/image2/plugin.js | 16 ++++++++++++++--
+ 1 file changed, 14 insertions(+), 2 deletions(-)
+
+diff --git a/plugins/image2/plugin.js b/plugins/image2/plugin.js
+index e85d58c2b..f40171644 100644
+--- a/plugins/image2/plugin.js
++++ b/plugins/image2/plugin.js
+@@ -375,8 +375,8 @@
+ 						hasCaption: !!this.parts.caption,
+ 						src: image.getAttribute( 'src' ),
+ 						alt: image.getAttribute( 'alt' ) || '',
+-						width: image.getAttribute( 'width' ) || '',
+-						height: image.getAttribute( 'height' ) || '',
++						width: image.$.style.width && parseInt( image.$.style.width ) || image.getAttribute( 'width' ) || '',
++						height: image.$.style.height && parseInt( image.$.style.height ) || image.getAttribute( 'height' ) || '',
+ 
+ 						// Lock ratio is on by default (https://dev.ckeditor.com/ticket/10833).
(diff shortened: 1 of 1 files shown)


patches/0010-LPS-139565-When-upgrading-from-6.2-to-7.1-image-widt.patch
```diff
@@ -0,0 +1,53 @@
...
++			}
++
++			var styleWidth = image.$.style.removeProperty('width');
++
++			if (styleWidth) {
++				styleWidth = styleWidth.replace('px', '');
++				widget.data.width = styleWidth;
++				image.$.setAttribute('width', styleWidth);
++			}
++
```
markocikos, line 36: While this code does the job, it is not placed in the ideal workflow spot. We are reading properties, as well as mutating underlying HTML in `onShow`, where I would not expect this logic to happen.

We can define [the internal data.width](https://github.com/ckeditor/ckeditor4/blob/master/plugins/image2/plugin.js#L378-L379) to read from `element.style`, with something like  
```js
data = {
	width: image.$.style.width && parseInt(image.$.style.width, 10) || image.getAttribute( 'width' ) || '',
	...
}
```
Then, we can cleanup HTML [on save](https://github.com/ckeditor/ckeditor4/blob/master/plugins/image2/plugin.js#L1117), with something like:
```js
if ( dimensions[ d ] )
	image.setAttribute( d, dimensions[ d ] );
	
	image.$.style.removeProperty(d);
```

patches/0010-LPS-139565-When-upgrading-from-6.2-to-7.1-image-widt.patch
```diff
@@ -0,0 +1,53 @@
...
+diff --git a/plugins/image2/plugin.js b/plugins/image2/plugin.js
+index e85d58c2b..59423cb79 100644
+--- a/plugins/image2/plugin.js
++++ b/plugins/image2/plugin.js
+@@ -1190,6 +1190,9 @@
+ 				nativeEvt, newWidth, newHeight, updateData,
+ 				moveDiffX, moveDiffY, moveRatio;
+ 
++			image.$.style.removeProperty('height');
++			image.$.style.removeProperty('width');
```
markocikos, line 49: Nice catch that we need to fix resize! :+1: 

The only issue here is there is a brief flash of image in its original size, before resize sets width and height attributes. I tried making a gif to show this, but the flash happens too quickly to be captured :slightly_smiling_face:   

To fix this, we just need to set width and height attributes here, before we remove the properties.

markocikos (changes requested):
@holatuwol Please see comments inline.

patches/0010-LPS-139565-When-upgrading-from-6.2-to-7.1-image-widt.patch
holatuwol: @markocikos I've applied the requested changes.

patches/0010-LPS-139565-When-upgrading-from-6.2-to-7.1-image-widt.patch
holatuwol: @markocikos I've applied the requested changes.

markocikos (approved):
LGTM!

PR #192, opened 2021-10-26
fix: makes codemirror and codemirrordialog plugin and dialog work with BalloonEditor

diegonvs:
# How to test it / Record

In order to test it, follow the provided steps at https://github.com/liferay-frontend/liferay-portal/pull/1612. Also there is a recording in this PR showing how it is working on DXP

# Additional comments

See comments on GH UI files


plugins/codemirror/dialogs/codemirrordialog.js
```diff
@@ -236,6 +236,10 @@ CKEDITOR.dialog.add('codemirrordialog', function (editor) {
 
 			if (newData !== oldData) {
 				editor.setData(newData);
+
+				if (editor.balloonToolbars) {
+					editor.balloonToolbars.hide();
+				}
```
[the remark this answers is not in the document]
julien: Thanks @diegonvs

markocikos:
@diegonvs The first commit should be `fix:` instead of `chore:`. I believe autogeneration of release docs ignores `chore`s. Also, to make CI happy, `fix: ` should be in PR title.

plugins/codemirror/dialogs/codemirrordialog.js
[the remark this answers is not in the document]
diegonvs: onLoad will not work because it is called just one time when the plugin is loaded. When moving to onShow it looks like a Tarsila do Amaral paint. Thanks Julien

diegonvs:
CI is happy and suggestions were squashed!

patches/0015-LPS-137763-If-contentsElement-is-defined-use-it-as-a.patch
```diff
@@ -0,0 +1,23 @@
...
+diff --git a/plugins/wysiwygarea/plugin.js b/plugins/wysiwygarea/plugin.js
+index 6c0a89d405..772839586c 100644
+--- a/plugins/wysiwygarea/plugin.js
++++ b/plugins/wysiwygarea/plugin.js
+@@ -42,7 +42,7 @@
+ 				iframe.setStyles( { width: '100%', height: '100%' } );
+ 				iframe.addClass( 'cke_wysiwyg_frame' ).addClass( 'cke_reset' );
+ 
+-				var contentSpace = editor.ui.space( 'contents' );
++				var contentSpace = editor.ui.space( 'contents' ) || editor.ui.contentsElement;
```
markocikos, line 20: There might be a bug here. What if both `editor.ui.space( 'contents' )` and `editor.ui.contentsElement` are set? Here, `editor.ui.space( 'contents' )` takes priority. But, in `themedui.js`, `editor.ui.contentsElement` takes priority:
```js
if (editor.ui.contentsElement) {
	editor.ui.contentsElement.setHtml('');
} else {
	editor.ui.space( 'contents' ).setHtml( '' );
}
```

Even if these two mutually exclusive, or duplicated, we should have consistent priorities for future-proofing.

plugins/codemirror/plugin.js
```diff
@@ -129,17 +126,22 @@
 				CKEDITOR.plugins.codemirror.commands.source
 			);
 
+			var codeMirrorDialogConfigs = {
+				state: CKEDITOR.TRISTATE_OFF,
+			};
+
+			if (!editor.balloonToolbars) {
+				codeMirrorDialogConfigs['modes'] = {
```
markocikos, line 134: Non-blocking, minor SF: If object property is a constant, we should use dot notation, `codeMirrorDialogConfigs.modes`. That way we don't declare a new string and gain a tiny bit on performance.

markocikos (commented):
@diegonvs  See comments inline.

patches/0015-LPS-137763-If-contentsElement-is-defined-use-it-as-a.patch
diegonvs: > we should have consistent priorities for future-proofing.

You're right. I changed themedui conditional to just use `editor.ui.contentsElement` in case if editor.ui.space is undefined ;)

plugins/codemirror/plugin.js
diegonvs: Makes sense. I'll change. I remember seeing this [pattern in linktoolbar](https://github.com/liferay/liferay-portal/blob/b6e202c32fb58fca5a54f9b4b5adb21351f67a42/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins/linktoolbar/plugin.js#L101)

diegonvs:
Ready for review again @markocikos

markocikos (approved):
LGTM!

markocikos (changes requested):
@diegonvs The code changes look good, but I just noticed we are adding two patches in the PR, 0014 and 0015. Can we squash these into one patch?

markocikos (approved):

PR #194, opened 2022-02-08
fix: make tableToolbar buttons work when we don't have a selection in table cells

diegonvs:
# How to reproduce this issue?

1. Open a Balloon Editor instance
2. Add a table using the insert/add button
3. When focusing in a cell, the tableToolbar will be shown
4. Try to click in one of the table buttons

We can't open the table options when clicking on table buttons and there are some errors on console caused by this issue

# Why you have done this?

I identified the issue is on [this LOC](https://github.com/ckeditor/ckeditor4/blob/a6dbe7467be13411cc6c85cb808cc7c10acab01c/plugins/panelbutton/plugin.js#L40) where `definition` is defined but we don't have `definition.panel`. In some cases, when `definition.panel` is undefined or just an empty object containing `{attributes: ''}`, we could use `definition` directly.

I tried to identify where we are passing this value of `definition` but before CKEditor code, we are defining it [at this part of](https://github.com/liferay/liferay-portal/blob/60362b2affa27e8a9499dd8c5b798fdbfc064e33/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins/tabletoolbar/plugin.js#L921) tableToolbar plugin.

The weird part here is because it just passes a key and some code inside the editor handles with finding the blockButton panel configs.

Screen Shot 2022-02-08 at 18 35 32

# Result

https://user-images.githubusercontent.com/7663513/153069108-dbaf4eb7-fcf1-4104-b0ba-4327b38c3015.mov


markocikos:
@diegonvs it looks like something went wrong in all the reverts and reverts of reverts. A part of my changes from https://github.com/liferay-frontend/liferay-portal/pull/1761 got reverted. [This code](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/editor/BalloonEditor.js#L88-L96) should not be in master. Text selection in table cell currently doesn't work. I thought we had a test of this, not sure how revert PR passed. This might be the real root cause of the issue, but it also could be unrelated. I'll see if I can bring back my changes, maybe that fixes this issue :crossed_fingers:

markocikos:
@diegonvs see https://github.com/liferay-frontend/liferay-portal/pull/1905

diegonvs:
> [This code](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/editor/BalloonEditor.js#L88-L96) should not be in master

Sorry 🙈 

I really don't know what happened why I (probably) changed this code while reverting-reverse-revertedsdsjbas :/

PR #202, opened 2022-10-25
fix: LPS-165655 Disable tab key by default to avoid the focus to be t…

victorg1991:
…rapped in the editor

Also, user can press CTRL + M to enable tabs in the editor.

This is the same solution we used to fix other similar cases: https://issues.liferay.com/browse/LPS-164875

### DIFF
--- a/plugins/codemirror/plugin.js
+++ b/plugins/codemirror/plugin.js
@@ -4,6 +4,47 @@
 	var stylesLoaded = false;
 
 	CKEDITOR.plugins.add('codemirror', {
+		_addTabKeyMessage: function (cm, editor, tabsKeyIsEnabled) {
+			var div = document.createElement('div');
+
+			var message = tabsKeyIsEnabled
+				? editor.lang.codemirror.disableTabKeyUsing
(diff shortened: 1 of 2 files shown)


plugins/codemirror/plugin.js
```diff
@@ -4,6 +4,47 @@
...
 	CKEDITOR.plugins.add('codemirror', {
+		_addTabKeyMessage: function (cm, editor, tabsKeyIsEnabled) {
+			var div = document.createElement('div');
+
+			var message = tabsKeyIsEnabled
+				? editor.lang.codemirror.disableTabKeyUsing
+				: editor.lang.codemirror.enableTabKeyUsing;
+
+			var html =
+				'<div class="keyboard-message popover px-2 py-1" style="left:auto; right: 4px; top:4px">';
```
dsanz, line 15: This introduces some clay-css dependencies, whick may not be available in all cases when using the editor in non-admin contexts. The bigger issue seems with the popover class as it defines the zindex and positioning. 

Probably with some inlining this can be worked around

PR #215, opened 2024-11-20
LPD-42437 Sets a default width in px rather than % for images with no width

antonio-ortega:
Hi,

A while ago, we fixed the case when images with no specified width (some SVG files, for example) where not rendered inside the editor. Our way to fix this, as you can see in [Patch 17](https://github.com/liferay/liferay-ckeditor/commit/9f4745ac63e9c7bf9ecfc95a971c6fa4cb0c0ef1), was to set `width:100%` for images.

However, that's causing some issues when we try to align the image as you can read in [LPD-42473](https://liferay.atlassian.net/browse/LPD-42473).

So instead of assigning a default width based on percentage I'm doing it based on a fix width, 150px, which might sound arbitrary but I took it from [this pull request](https://github.com/ckeditor/ckeditor4/pull/2668), which never got committed because the sender didn't make changes requested but the idea of a fix 150px width for this kind of images seemed ok to ckeditor's developers.

Let me know if you have further questions.

Thanks.


markocikos (approved):
@antonio-ortega LGTM! :sparkles: 

Just a sidenote: you don't need to rebuild artifacts with each PR. We can just do it once before release, as a part of release.

PR #216, opened 2024-12-04
LPD-42473 Instead of setting a width from image2 plugin, we'll do it in portal

antonio-ortega:
Hi,

This new patch (#21) is basically reverting our previous tries (#20, #17) to render images with no width. 

After a deeper analysis, we've seen none of these solutions is working 100% and this issue should be better handled inside Liferay Portal, to be more exact, within [itemselector](https://github.com/liferay/liferay-portal/blob/master/modules/apps/frontend-editor/frontend-editor-ckeditor-web/src/main/resources/META-INF/resources/_diffs/plugins/itemselector/plugin.js#L248-L288) plugin.

Once this pr gets merged, I'll release a new liferay-ckeditor version, and then I'll send a pull request to liferay-portal upgrade such version and including the fix within itemselector plugin.

Thanks.

Regards.

### DIFF
--- a/patches/0021-LPD-42473-Instead-of-setting-a-width-from-image2-plu.patch
+++ b/patches/0021-LPD-42473-Instead-of-setting-a-width-from-image2-plu.patch
@@ -1,0 +1,25 @@
+From f49502373d3f8115adfe6184a921ca9a4b26ebe7 Mon Sep 17 00:00:00 2001
+From: Antonio Ortega <60252917@liferay.com>
+Date: Wed, 4 Dec 2024 14:09:03 +0100
+Subject: [PATCH] LPD-42473 Instead of setting a width from image2 plugin,
+ we'll do it directly inside portal when rendering the image
+
+---
+ plugins/image2/plugin.js | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/plugins/image2/plugin.js b/plugins/image2/plugin.js
+index da58812eff..bcf986bad4 100644
+--- a/plugins/image2/plugin.js
++++ b/plugins/image2/plugin.js
+@@ -376,8 +376,8 @@
+ 						hasCaption: !!this.parts.caption,
+ 						src: image.getAttribute( 'src' ),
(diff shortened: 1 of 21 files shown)


patches/0021-LPD-42473-Instead-of-setting-a-width-from-image2-plu.patch
```diff
@@ -0,0 +1,25 @@
...
+--- a/plugins/image2/plugin.js
++++ b/plugins/image2/plugin.js
+@@ -376,8 +376,8 @@
+ 						hasCaption: !!this.parts.caption,
+ 						src: image.getAttribute( 'src' ),
+ 						alt: image.getAttribute( 'alt' ) || '',
+-						width: image.$.style.width && parseInt( image.$.style.width ) || image.getAttribute( 'width' ) || '150px',
+-						height: image.$.style.height && parseInt( image.$.style.height ) || image.getAttribute( 'height' ) || 'auto',
++						width: image.$.style.width && parseInt( image.$.style.width ) || image.getAttribute( 'width' ) || '',
++						height: image.$.style.height && parseInt( image.$.style.height ) || image.getAttribute( 'height' ) || '',
```
markocikos, line 22: There is a line deleted in patch 20 that we're not reverting,
```js
'.cke_widget_wrapper.cke_widget_image {width: 100%;}'
```
Is this intentional?

If it is, this PR is fine as it is. If it is a full revert, can we just delete patch 20 and rebuild?

markocikos (commented):
@antonio-ortega A question inline.

patches/0021-LPD-42473-Instead-of-setting-a-width-from-image2-plu.patch
antonio-ortega: It was not intentional :( Thanks for catching it!

antonio-ortega: Yes, because that line was added in patch 17... then removed in patch 20 by this new solution than I'm also reverting now.

Basically I'm getting rid of our tries to get this behavior inside liferay-ckeditor. We'll do it in portal.

PR #218, opened 2025-01-10
LPD-45795 Check element exists

fortunatomaldonado:
https://liferay.atlassian.net/browse/LPD-45795

When an image tag is surrounded by several span tags, CKEditor cannot find the widget element the image belongs too and causes an error.

In order to prevent this, we need to check if the element exists before continuing.

Please let me know if there are any questions or comments about this.
Thank you!


bryceosterhaus (commented):
The changes to check element seem fine to me. All of the other changes seem odd, the ones that change all the hashes. But I'm also unfamiliar with ckeditor updates. maybe @markocikos can confirm this is all good

markocikos (approved):

@bryceosterhaus The hash changes are output of automatic rebuild of ckeditor from dev to production minified version. @fortunatomaldonado it's not wrong, but we don't need to do this on every PR. It can be done once, as a part of release.

Speaking of release, @fortunatomaldonado do you still have permission issues when making a release?

markocikos:
@fortunatomaldonado This time you don't have to do the release, I'll do it to after fix for [LPP-57356](https://liferay.atlassian.net/browse/LPP-57356)

PR #221, opened 2025-01-20
LPD-45919 Remove plugins/codesnippetgeshi/dev directory

antonio-ortega:
Hi,

Since https://nvd.nist.gov/vuln/detail/cve-2024-43407 was reported, I'm following [the same approach](https://github.com/ckeditor/ckeditor4/commit/71072c9f7f263329841bd38e7e5309074c82ef94) CKEditor did to fix it removing codesnippetgeshi/dev directory.

Thanks.

PR #220, opened 2025-01-14
LPD-45919 Remove CKEditor 4 samples from production build

markocikos:
### References

- JIRA: [LPD-45919 Remove CKEditor 4 samples from production build](https://issues.liferay.com/browse/LPD-45919)

### What is the goal of this PR?

The goal of the task is to remove all internal CKEditor 4 samples. The only relevant samples that we want to maintain are the ones in [frontend-editor-ckeditor-sample-web](https://github.com/liferay/liferay-portal/tree/master/modules/apps/frontend-editor/frontend-editor-ckeditor-sample-web), covered by tests.

I’m opening this ticket as a bug, because we are technically fixing XSS vulnerability in one of the samples. 

This task will result is a slight performance boost.

### What does it look like?

image

image

image

PR #222, opened 2025-01-21
LPD-45919 Remove plugins/codesnippetgeshi so it is not shipped in our liferay-ckeditor

antonio-ortega:
[Previous pull requeest](https://github.com/liferay/liferay-ckeditor/pull/221)

As suggested [here](https://github.com/liferay/liferay-ckeditor/pull/221#discussion_r1923401242) rather than creating a huge patch, we can do the same just removing that directory when building liferay-ckeditor.

### DIFF
--- a/ck.sh
+++ b/ck.sh
@@ -69,6 +69,9 @@ case "$COMMAND" in
 				rm -rf temp
 				cp -r ckeditor-dev temp
 
+				# Remove codesnippetgeshi plugin
+				rm -rf temp/plugins/codesnippetgeshi
+
 				# Copy custom skins
 				cp -r skins temp
 


ck.sh
```diff
@@ -69,6 +69,9 @@ case "$COMMAND" in
 				rm -rf temp
 				cp -r ckeditor-dev temp
 
+				# Remove codesnippetgeshi plugin
+				rm -rf temp/plugins/codesnippetgeshi
+
```
markocikos, line 74: This is inventive, but not exactly what I had in mind. With this, we would first make a production build, that includes unwanted dir, and delete that dir after. A better solution is to exclude unwanted dir from production build itself, so it's not added in the first place. We can do this by adding exclusion pattern to [CKBUILDER_CONFIG in build-config.js](https://github.com/liferay/liferay-ckeditor/blob/master/support/build-config.js#L10-L34). That's what I did in https://github.com/liferay/liferay-ckeditor/pull/220, adding `samples` string to `ignore` array. Note that there is a version of this file in both dev subrepo and main repo, as the build copies the config file itself. It may be safe to delete file in main repo, but I wasn't sure, so I just updated it in both places.

Is the intent to remove codesnippetgeshi/dev dir, or the whole plugin? If it's dev dir, I am noticing `dev` is [already in the exclusion list](https://github.com/liferay/liferay-ckeditor/blob/master/support/build-config.js#L16), and codesnippetgeshi/dev is already not added to [production build](https://github.com/liferay/liferay-ckeditor/tree/master/ckeditor/plugins/codesnippetgeshi). If the intent is to remove the whole plugin, just adding `codesnippetgeshi` to exclusion list should do it. Is this safe to remove the whole plugin?

ckeditor-dev
markocikos, line 1: I am not sure what generated this change, but it looks wrong and is probably causing the CI failure.

markocikos (changes requested):
@antonio-ortega please see comments inside

ck.sh
antonio-ortega: That's a really good point! I hadn't noticed we are already removing `dev` folder from plugin, which is exactly the way how CKEditor [addressed](https://github.com/ckeditor/ckeditor4/commit/71072c9f7f263329841bd38e7e5309074c82ef94) this case. 

Having a look at their current master branch, where this is issue is considered fixed, they still have that [plugin.js](https://github.com/ckeditor/ckeditor4/tree/master/plugins/codesnippetgeshi) file.

Based on that, I think we are not vulnerable to this issue and we can close this pull request.

Thanks.

Regards.

PR #224, opened 2025-05-02
LPD-54093 Remove IE11 fixes causing focus issues with scrollbar

fortunatomaldonado:
We need to include https://github.com/fortunatomaldonado/liferay-portal/commit/733b048d8f82bcef89a6283bd5f55ced540f3007 first before we reproduce.

Steps to reproduce:
1. Using Chrome, Create a new Object (e.g. TestObject)
2. Add a single Rich text field to it
3. Publish the Object
4. Navigate to Control Panel > Applications > Process Builder and enable the Single Approver workflow for your Object.
5. Add a new page and place your Object on it.
6. Add a new Object entry by copying and pasting the entire content in the attached file (rich_text_content.doc), notice you can use the scrollbar on the bottom of the rich text field to scroll left and right.
7. Click on Save
8. Open the User Icon > Notifications and click on the Review link
9. Dont touch the Editor. Scroll down to the end of the editor and move the horizontal scrollbar.

Actual results: The focus move up
Expected results: Focus is not moved

This IE11 fix would trigger and cause the focus to jump to the editor.
It looks like removing this IE11 fix does not affect any other aspect and should be safe.

https://github.com/user-attachments/assets/4bb9fb69-24f5-4d46-b32c-cef03a9aaa0f


patches/0025-LPD-54093-Remove-IE11-fixes-causing-focus-issue-with.patch
```diff
@@ -0,0 +1,34 @@
...
+index 97c849a80c..45bc136008 100644
+--- a/plugins/wysiwygarea/plugin.js
++++ b/plugins/wysiwygarea/plugin.js
+@@ -268,19 +268,6 @@
+ 			removeSuperfluousElement( 'div' );
+ 		}
+ 
+-		// Fix problem with cursor not appearing in Webkit and IE11+ when clicking below the body (https://dev.ckeditor.com/ticket/10945, https://dev.ckeditor.com/ticket/10906).
+-		// Fix for older IEs (8-10 and QM) is placed inside selection.js.
+-		if ( CKEDITOR.env.webkit || ( CKEDITOR.env.ie && CKEDITOR.env.version > 10 ) ) {
```
markocikos, line 21: Issue title and the patch name confused me a bit, since IE11 fix should be behind environment `if` check, and not have any effect on code. The key bit here seems to be `CKEDITOR.env.webkit`, which applies the code to Safari and... Chrome on iOS and Linux? I thought Chrome does not use webkit, but apparently it does. My Chrome on Linux definitely does return true for `CKEDITOR.env.webkit`.

So, we can push this code as it is, but I'd make two small changes:
- Just remove `CKEDITOR.env.webkit` check, leaving fix for IE. Yeah, we don't care about IE, but why not leave in perfectly good fix when we already have it.
- I'd name patch something like `Do-not-apply-focus-fix-to-webkit`

markocikos (commented):
@fortunatomaldonado Please see comment inline

markocikos (approved):

PR #225, opened 2025-09-15
LPD-63066 Improve shortcut hint

antonio-ortega:
Hi,

This is related to LPD-63066 which I initially solved [this way](https://github.com/liferay-frontend/liferay-portal/pull/5071), but then @marcoscv-work warned my solution might cause an accessibility issue because I was not showing the initial popover for a few seconds before being able to minimize/maximize by hovering on it.

I realized that solution was easier to implement in liferay-ckeditor rather than liferay-portal, so once we update the version in portal I'll revert previous commit related to this LPD.

Thanks.

Regards.


plugins/codemirror/plugin.js
```diff
@@ -35,6 +35,13 @@
 			div.innerHTML = html;
 
 			cm.getWrapperElement().appendChild(div);
+
+			setTimeout(function () {
+				var popover = div.querySelector('.popover');
+				if (popover) {
+					popover.classList.add('d-reduce');
+				}
+			}, 4000);
```
markocikos, line 44: Seems a bit arbitrary, but if PE is ok with this, :+1:

ckeditor-dev
markocikos, line 1: I'm not sure, what happened here, but there should be no changes on subrepo with this PR.

markocikos (changes requested):
@antonio-ortega apart for build issues, LGTM.

antonio-ortega:
Resent https://github.com/liferay/liferay-ckeditor/pull/226

PR #228, opened 2026-02-18
LPD-79554 Always update content on change

fortunatomaldonado:
https://liferay.atlassian.net/browse/LPD-79554
https://liferay.atlassian.net/browse/LPP-62862

Issue: When editing in source mode, the content is not always adjusted when going out of source mode.

This is happening since we are only changing the content if the old data is different than the new data. 

EX:
1. we start with `hello` in content
2. Switch to source and write `hello1`
3. Erase the 1 leaving you with `hello`
4. Switch out of source mode

Actual Behavior: We see `hello1`

Expected Behavior: We see `hello`

To solve this we should always update the with the new data.

Let me know if there are any questions or comments about this.
Thank you!


markocikos (approved):
@fortunatomaldonado LGTM! ✨

